PR #22, opened 2015-08-27
Provide deployment into vagrant box. Closes #21.

### LINKED ISSUE #21: Deploy into Vagrant
For a development installation, it would be convenient if the installation procedure was automated. For that, Vagrant could be used, resulting in a single VM that runs both the application and Docker.

loewis:


jprberlin:
lgtm. thx,

PR #37, opened 2016-01-19
instead of relying on shared folders, use docker api archive function…

rteusner:
…. un tar is still a bit cumbersome, but works.


tstaubitz:
lgtm +1

PR #39, opened 2016-01-25
Render command

rteusner:


rteusner:
lgtm

rteusner:
No longer needed, was done in the normal output / write command.

PR #41, opened 2016-02-04
Import exercise meta data from CodeHarbour

leoselig:
Authentication via
- OAuth2 tokens
- configured in code harbor links (new mvc)

Import via
- ProFormA XML

Caveats:
- imports only meta data (title, description) right now
- associates with execution environment with id 1 (to pass validation)


app/controllers/exercises_controller.rb
```diff
@@ -11,6 +11,10 @@ class ExercisesController < ApplicationController
   before_action :set_file_types, only: [:create, :edit, :new, :update]
   before_action :set_teams, only: [:create, :edit, :new, :update]
 
+  skip_before_filter :verify_authenticity_token, only: [:import_proforma_xml]
```
leoselig: add code comment clarifying why we are skipping Pundit authorization here

PR #47, opened 2016-03-10
Testruns

Parnswir:
This persists and visualizes unit test results and shows simple error messages instead of the whole stack trace.

Error messages are parsed by the test framework adapter and are optional. Currently there are implementations for Java and Python.


rteusner:
lgtm

PR #53, opened 2016-05-10
Fix file creation behavior

Parnswir:
This pull request allows to define a boolean flag "allow file creation" for exercises and fixes the creation of files with the same name in the editor.


rteusner:
lgtm

PR #57, opened 2016-05-12
Percentage instead score

yqbk:


rteusner:
lgtm

rteusner:
Please add some rounding down, so that we don't have fractions

rteusner:
lgtm, merge.

PR #64, opened 2016-06-20
Score websocket

rteusner:


app/assets/javascripts/editor.js.erb
```diff
@@ -184,8 +185,8 @@ $(function() {
     (streamed ? evaluateCodeWithStreamedResponse : evaluateCodeWithoutStreamedResponse)(url, callback);
   };
```
tstaubitz: Please remove server side event logic completely.

lib/docker_client.rb
```diff
@@ -200,7 +200,7 @@ def execute_arbitrary_command(command, &block)
     execute_command(command, nil, block)
   end
 
-  #only used by server sent events (deprecated?)
+  #only used by score
   def execute_command(command, before_execution_block, output_consuming_block)
```
tstaubitz: Have a look whether score calls execute_command

tstaubitz:
apart from that 👍

lib/docker_client.rb
rteusner: I had a look.

PR #66, opened 2016-06-20
Feature file templates

rteusner:


app/policies/file_template_policy.rb
```diff
@@ -0,0 +1,11 @@
+class FileTemplatePolicy < AdminOnlyPolicy
```
jprberlin: Why is this inherited from adminpolicy if everyone is allowed

rteusner: Because everyone can see them, but only admins can create or update them.

PR #72, opened 2016-07-07
Rfc cleanup

rteusner:


app/models/request_for_comment.rb
```diff
@@ -25,6 +25,35 @@ def last_submission
...
+            '#{created_at.strftime('%Y-%m-%d %H:%M:%S.%N')}' > created_at
+            order by created_at desc
+            limit 1").first
+    submission4 = Submission.find_by_sql(" select * from submissions
+            where exercise_id = #{exercise_id} AND
+            user_id =  #{user_id} AND
+            '#{created_at.localtime.strftime('%Y-%m-%d %H:%M:%S.%N')}' > created_at
+            order by created_at desc
+            limit 1").first
+    binding.pry
```
tstaubitz: shouldn't that be removed for production?

PR #73, opened 2016-07-11
Code editor features

yqbk:
need to provide source for ext-language_tools.js (instead of github link)


rteusner:
Moved to the editor-frontend-refactor branch, also added additional checkbox for exercise there.

PR #84, opened 2016-12-30
Lti param model

tstaubitz:
This branch replaces the storage of the LtiParameters in a session with storing these parameters in a proper model.

It also adds a semi-automated installer to run CodeOcean on a "Non-Vagrant" local VM (Debian Jessie)


spec/factories/lti_parameter.rb
```diff
@@ -1,12 +1,23 @@
 FactoryGirl.define do
+  #TODO do this proper
   factory :lti_parameter do
-    association :consumers_id, factory: :consumer
-    association :exercises_id, factory: :math
-    association :external_users_id, factory: :external_user
+    consumers_id 1
+    exercises_id 1
```
rteusner: why did you remove association here and set constant values?

app/controllers/exercises_controller.rb
```diff
@@ -158,9 +158,11 @@ def index
 
   def redirect_to_lti_return_path
     lti_parameter = LtiParameter.where(consumers_id: session[:consumer_id],
-                                       external_users_id: @current_user.id,
+                                       external_users_id: @submission.user_id,
                                        exercises_id: @submission.exercise_id).first
 
+    # binding.pry
```
rteusner: remove

PR #89, opened 2017-02-14
Pullrequest proxy exercises

ThommyH:
Extended Exercises by worktime, difficulty and tags
Tags can be added to exercises in the edit view. Tags can monitored under /tags.
Added the concept of ProxyExercises which are a collection of Exercises. They can be found under /proxy_exercises
Added Interventions as prework to show interventions later to the user.
Added exercise/[:id]/working_time to return the working time of the user in this exercise and the average working time of all users in this exercise


app/assets/remote_scripts/client_script.ps1
```diff
@@ -0,0 +1,90 @@
+# run like this: powershell.exe -noprofile -executionpolicy bypass -file path\to\client_script.ps1 path\to\project_root
```
ThommyH: I have no idea why this pops up here, was this file removed recently?

PR #118, opened 2017-06-09
Rfc mark as answered extensions

Parnswir:
Adds a small dialog to send a thank you note to all users who helped answer a request for comments.


rteusner:
lgtm

PR #120, opened 2017-06-23
Close flash messages

Parnswir:
This adds a button to close a flash message.


lib/assets/stylesheets/flash.css.scss
```diff
@@ -13,14 +20,4 @@
   padding-top: 0;
 }
 
-.clickthrough {
```
rteusner: what about leaving this, for reference reasons? (I would like this and it would not take too much space, what do you think?)

rteusner:
lgtm

PR #125, opened 2017-09-06
Rework comment modal

Parnswir:


Parnswir:
hi @rteusner, using the annotation property provided by Ace sets the icons in the gutter area and provides basic tooltips (only one string per annotation). At runtime it proved unsuitable to use the icon elements as event receivers for the bootstrap popovers we now use to display detailed information about the comment history. Since the clustering is done only once per "refresh cycle", it does not need to recompute clusters after the UI is built once. Also, we only need to build popover and popup UI elements for lines which do have comments.

Parnswir:
@rteusner Please have a look at the current status. From my POV, these features should be ready to be merged. Included are:
* detailed comment history in popup
* resizeable/draggable popup
* max. 3 comments in popover (preview)
* edit/delete actions for comments of the current user
* subscription feature for new comments of the author of the RFC (subscribe/unsubscribe via UI and email)
* new extensible backend model for subscriptions (not only author's comments in the future)

Parnswir:
@rteusner I have rebuilt the popover creation from scratch based on the mouseover event of the ace-gutter. There is some painful "de-bouncing" of events involved, but overall this is much cleaner and fixes the scrolling and folding problems of the previous implementation.

rteusner:
lgtm 👍

PR #126, opened 2017-09-20
Better Subscriptions

Parnswir:
- subscriptions are insert-only; if `deleted`==`false`, `updated_at` is the deletion date.
- there will only be one notification per user, even if there is more than one subscription for this user
- the subscription box is now always visible (not only if someone commented before opening the modal)


rteusner:
lgtm

PR #129, opened 2017-10-03
removed expected working time from show/create/index of exercises

ThommyH:
we do not use it


db/schema.rb
```diff
@@ -47,6 +47,13 @@
     t.string   "oauth_secret", limit: 255
   end
 
+  create_table "copy_paste_events", id: false, force: :cascade do |t|
```
Parnswir: Most of the schema changes are unrelated to the working time.

ThommyH:
removed them

PR #132, opened 2017-10-04
Updated ProFormA-XML import

Adrian-St:


app/models/exercise.rb
```diff
@@ -300,35 +366,108 @@ def determine_file_role_from_proforma_file(task_node, file_node)
...
+  def build_proforma_xml_for_test(xml, test, index)
+    proforma = xml['p']
+    proforma.test('id' => 't' + index.to_s) {
+      proforma.title('')
+      proforma.send('test-type', 'unittest')
+      proforma.send('test-configuration') {
+        proforma.filerefs {
+          proforma.fileref('refid' => test.id.to_s)
+        }
+        xml['u'].unittest('framework' => 'JUnit 4', 'version' => '')
```
tstaubitz: You need to check here for which programming language the exercise is. If it is Java, then use JUnit 4 (split frame work and version), if it is Python then use Pyunit, if it is Ruby, then use RSpec. We intentionally decided to leave this hard-coded in here as everything else would require very high efforts, which do not bring us any further. So the check for the programming language is not a nice solution but the minimum viable product.

app/models/exercise.rb
```diff
@@ -300,35 +366,108 @@ def determine_file_role_from_proforma_file(task_node, file_node)
...
+        proforma.fileref('refid' => model_solution_file.id.to_s)
+      }
     }
-    self.execution_environment_id = 1
+  end
+
+  def to_proforma_xml
+    builder = Nokogiri::XML::Builder.new do |xml|
+      proforma = xml['p']
+      proforma.task('xmlns:p' => 'urn:proforma:task:v1.1', 'lang' => 'en', 'uuid' => SecureRandom.uuid,
```
tstaubitz: change the default language here to German, please. Most of our exercises are German, currently, so it will be less work to adjust the others.

app/policies/exercise_policy.rb
```diff
@@ -12,7 +12,7 @@ def show?
     @user.internal_user?
   end
 
-  [:clone?, :destroy?, :edit?, :statistics?, :update?, :feedback?].each do |action|
```
tstaubitz: Why did you delete the feedback here? We suspect this went wrong during the merge?

app/controllers/exercises_controller.rb
```diff
@@ -93,6 +97,17 @@ def edit
     collect_set_and_unset_exercise_tags
   end
 
+  def push_proforma_xml
+    codeharbor_link = CodeHarborLink.find(params[:account_link])
+    oauth2Client = OAuth2::Client.new('client_id', 'client_secret', :site => codeharbor_link.push_url)
```
tstaubitz: client_id and client_secret will be set correctly in the next merge_request.
We will adjust the procedure on CodeHarbor accordingly. 
A group admin will be able to register an autograder with CodeHarbor.
The groups on CodeHarbor will be extended with a third role: instructor

- admins can add and edit users, exercises and autograders to the group.  
- members can view the exercises and export them to xml
- instructors can do the same as members plus pushing them to the registered autograder.

tstaubitz:
@rteusner We just decided that there will be no automatic mapping from programming language (CodeHarbor) to execution environment (CodeOcean). There are too many options. Basically the execution_environment needs to be manually selected.

The problem is however, that we need an execution_environment when an exercise is created.
Solution 1: we make the execution_environment optional   
Solution 2: we add a default execution_environment  in the database with some sort of basic docker container that actually does nothing as a placeholder until the user selects a proper environment. 

A hint will be shown to the teacher either way that some sort of action is required both in the exercise list and the exercise show/edit view.

My preference would be Solution 1 as in the end the outcome is the same and the solution is less error prone and way more straight forward. 

Opinions?

app/views/exercises/show.html.slim
```diff
@@ -7,10 +7,19 @@ h1
...
 
+
 = row(label: 'exercise.title', value: @exercise.title)
 = row(label: 'exercise.user', value: link_to_if(policy(@exercise.author).show?, @exercise.author, @exercise.author))
 = row(label: 'exercise.description', value: render_markdown(@exercise.description))
-= row(label: 'exercise.execution_environment', value: link_to_if(policy(@exercise.execution_environment).show?, @exercise.execution_environment, @exercise.execution_environment))
+= row(label: 'exercise.execution_environment') do
+  = link_to_if(policy(@exercise.execution_environment).show?, @exercise.execution_environment, @exercise.execution_environment)
+  - if @exercise.default_execution_environment
+    .default-execution-environment-warning
```
rteusner: default execution environment warning? Is that specifically for imported exercises?

config/docker.yml.erb.travis
```diff
@@ -7,7 +7,7 @@ default: &default
 
 development:
   <<: *default
-  host: tcp://127.0.0.1:2376
+  host: tcp://0.0.0.0:2376
```
rteusner: is this necessary? Travis seems to be working atm, why change it?

db/migrate/20180312135814_add_default_execution_environment_to_exercises.rb
```diff
@@ -0,0 +1,5 @@
+class AddDefaultExecutionEnvironmentToExercises < ActiveRecord::Migration
+  def change
+    add_column :exercises, :default_execution_environment, :boolean, :default => false
```
rteusner: how is that intended to work?

db/schema.rb
```diff
@@ -123,18 +124,19 @@
...
-    t.string   "title",                    limit: 255
+    t.string   "title",                         limit: 255
     t.datetime "created_at"
     t.datetime "updated_at"
     t.integer  "user_id"
     t.text     "instructions"
     t.boolean  "public"
-    t.string   "user_type",                limit: 255
-    t.string   "token",                    limit: 255
+    t.string   "user_type",                     limit: 255
```
rteusner: clear those changes please...

debian_installer/setup_debian_1_install_postgres.sh
```diff
@@ -11,7 +11,7 @@ then
   echo "Add Postgres sources..."
   cd /etc/apt/sources.list.d
   sudo touch pgdg.list
```
rteusner: these changes do not belong into this pull request.

lib/proforma/xml_generator.rb
```diff
@@ -0,0 +1,142 @@
+require 'nokogiri'
```
rteusner: this serves as the "exporter" ?

db/migrate/20180312135814_add_default_execution_environment_to_exercises.rb
rteusner: I kinda dislike that solution...

app/assets/javascripts/execution_environments.js
```diff
@@ -1,7 +1,137 @@
...
   if ($.isController('execution_environments')) {
     if ($('.edit_execution_environment, .new_execution_environment').isPresent()) {
       new MarkdownEditor('#execution_environment_help');
+      setup_programming_language();
+      //show_programming_language_values();
+      setup_chosen_fields();
     }
   }
 });
+
```
rteusner: this is called also outside of the view / controller of execution_environments.... I suppose this is not intended?

If it is - why?

app/controllers/exercises_controller.rb
rteusner: this has not yet happened.

app/policies/exercise_policy.rb
rteusner: We still need an answer here

app/policies/exercise_policy.rb
```diff
@@ -12,7 +12,7 @@ def show?
...
+  [:clone?, :destroy?, :edit?, :statistics?, :update?, :push_proforma_xml?, :feedback?].each do |action|
```
rteusner: ah it was re-added here...

app/policies/programming_language_policy.rb
```diff
@@ -0,0 +1,5 @@
+class ProgrammingLanguagePolicy < AdminOrAuthorPolicy
+  [:create?, :versions?].each do |action|
```
rteusner: please unnest.

app/views/exercises/index.html.slim
```diff
@@ -28,7 +28,15 @@ h1 = Exercise.model_name.human(count: 2)
       - @exercises.each do |exercise|
         tr data-id=exercise.id
           td = exercise.title
-          td = link_to_if(exercise.execution_environment && policy(exercise.execution_environment).show?, exercise.execution_environment, exercise.execution_environment)
+          td
+            - if exercise.execution_environment
+              = link_to_if(policy(exercise.execution_environment).show?, exercise.execution_environment, exercise.execution_environment)
+            - else
+              i.fa.fa-warning style='color: zred;'
```
Parnswir: Avoid inline styles. Also: typo?

config/locales/de.yml
```diff
@@ -620,6 +648,7 @@ de:
     upload_file: Datei hochladen
     time:
       before: "%{time} zuvor"
+    "yes": Ja
```
rteusner: Does this need to be in quotation marks?

app/views/exercises/show.html.slim
```diff
@@ -7,10 +7,20 @@ h1
   - if policy(@exercise).edit?
     = render('shared/edit_button', object: @exercise)
 
+
 = row(label: 'exercise.title', value: @exercise.title)
 = row(label: 'exercise.user', value: link_to_if(policy(@exercise.author).show?, @exercise.author, @exercise.author))
 = row(label: 'exercise.description', value: render_markdown(@exercise.description))
-= row(label: 'exercise.execution_environment', value: link_to_if(policy(@exercise.execution_environment).show?, @exercise.execution_environment, @exercise.execution_environment))
+= row(label: 'exercise.execution_environment') do
+  - if @exercise.execution_environment
```
Parnswir: This block looks nearly identical to https://github.com/openHPI/codeocean/pull/132/files#diff-65b4469e16350935fe943bda56060ae2R32 -- it might be good to extract the template and `render` it here as a component

lib/tasks/export_exercise_collection.rake
```diff
@@ -0,0 +1,19 @@
+namespace :codeharbor do
+
+  desc 'write displaynames retrieved from the account service as csv into the codeocean database'
```
rteusner: this description seems to be the leftover of a copy and paste...

lib/tasks/export_exercise_collection.rake
```diff
@@ -0,0 +1,19 @@
...
+
+  task :export_exercise_collection, [:exercise_collection_id, :push_url, :push_token] => [:environment] do |t, args|
+
+    begin
+      ExerciseCollection.find(args.exercise_collection_id).exercises.each do |exercise|
+        oauth2Client = OAuth2::Client.new('client_id', 'client_secret', :site => args.push_url)
```
rteusner: still hard coded strings here

test/models/programming_language_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
```
rteusner: this seems to be leftover? --> Remove

test/models/programming_languages_join_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
```
rteusner: Same here: Leftover? -> Remove.

app/views/execution_environments/_programming_languages_join_fields.html.slim
```diff
@@ -0,0 +1,20 @@
...
+  = f.hidden_field :_destroy
+  - f.object.build_programming_language unless f.object.programming_language
+  = f.fields_for :programming_language do |ff|
+    .col-md-4.name
+      = ff.object.name
+    .col-md-4.version
+      = ff.object.version
+    .col-md-3.default
+      span.true style="display:#{display_true}"
+        = t('shared.yes')
```
rteusner: please use icons here. Remove the shared strings yes and no.

app/views/execution_environments/_programming_languages_join_fields.html.slim
```diff
@@ -0,0 +1,20 @@
...
+      span.false style="display:#{display_false}"
+        = t('shared.no')
```
rteusner: same here

app/assets/javascripts/execution_environments.js
Adrian-St: I don't really understand what you mean.
setup_programming_language() and setup_chosen_fields() is only supposed to be called when loading the form.
But I thought that's the point of these lines: 

if ($.isController('execution_environments')) {
    if ($('.edit_execution_environment, .new_execution_environment').isPresent())

Am I mistaken? If so, how do I make sure it is only called with the form?

lib/proforma/xml_generator.rb
Adrian-St: Yes, nokogiri is a gem to generate and read xml data.

app/assets/javascripts/execution_environments.js
Parnswir: I think the comment refers to the block below, lines 95-137. 😄

config/locales/de.yml
Parnswir: `yes` and `no` are key words in YAML, they need to be in quotes, because they would otherwise evaluate to `true` and `false` [at runtime](https://makandracards.com/makandra/24809-yaml-keys-like-yes-or-no-evaluate-to-true-and-false).

app/assets/javascripts/execution_environments.js
Adrian-St: Yeah, I saw it now, has been fixed :D

lib/tasks/export_exercise_collection.rake
Adrian-St: I think they are okay here, since we only use rake tasks internally and client_id and client_secret are only there because they are required by OAuth2, but are not used otherwise.
What do you think?

MrSerth:
superseded by #423

PR #133, opened 2017-10-04
allow only one intervention per user and exercise but 3 interventions per user and day

ThommyH:
users could get unlimited amount of interventions per exercise if they split it to multiple days. We agreed to limit it to one intervention per exercise but 3 per day in total.


app/controllers/exercises_controller.rb
```diff
@@ -166,17 +170,19 @@ def handle_file_uploads
...
+    user_got_enough_interventions = count_interventions_today >= max_intervention_count_per_day || user_got_intervention_in_exercise
     is_java_course = @course_token && @course_token.eql?(java_course_token)
 
     user_intervention_group = UserGroupSeparator.getInterventionGroup(current_user)
 
     case user_intervention_group
       when :no_intervention
       when :break_intervention
-        @show_break_interventions = (user_solved_exercise || !is_java_course || user_got_enough_interventions) ? "false" : "true"
+        @show_break_interventions = (!user_solved_exercise && is_java_course && !user_got_enough_interventions) ? "true" : "false"
```
ThommyH: greetings from demorgan :D

Parnswir: Imho you could leave out the conditional, i.e. 
```ruby
@show_break_interventions = not user_solved_exercise and is_java_course and not user_got_enough_interventions
```
Looks a bit more readable 😄

PR #141, opened 2017-11-22
Update exercise feedback UI

Parnswir:


db/schema.rb
```diff
@@ -104,6 +104,7 @@
     t.string   "name"
     t.datetime "created_at"
     t.datetime "updated_at"
+    t.boolean  "use_anomaly_detection", default: false
```
rteusner: does not seem to belong here ;)

PR #150, opened 2018-02-22
Fix feedback timestamps

rteusner:
fixed several severe bugs, that were still in the version update-gemfile.
One resulted in a segmentation fault on debian, the other one prevented lti sign-ins.
deployment was also broken and was fixed with an recursive flag.

And we fixed the feedback timestamps (removed default that overwrote correct timestamps)


MrSerth:
:shipit:

PR #153, opened 2018-02-28
Migration fixes

Parnswir:
* migrate RFC-attributes in batches of 1000 to [save memory](https://rubyinrails.com/2017/11/16/use-find-each-instead-of-all-each-in-rails/)
* make migrations (at least pseudo-) reversible


rteusner:
lgtm

PR #156, opened 2018-03-07
Grow/shrink buttons depending on number of buttons visible

Parnswir:


rteusner:
cool fix, thank you :)

PR #159, opened 2018-03-09
Update Vagrant provision.sh to execute command as non-root user

MrSerth:
Also, some commands were only available in an interactive shell, those were changed.


rteusner:
lgtm

PR #162, opened 2018-03-21
Fix statistics for empty collections and exercises without submissions

Parnswir:


rteusner:
fix looks good :)

rteusner:
however, it has still an error:
Error message
NoMethodError: undefined method `/' for nil:NilClass

Sample stack trace (show Rails)
…eases/20180322141236/app/models/exercise_collection.rb:  20:in `average_working_time'
…3.0/gems/actionview-4.2.10/lib/action_view/template.rb: 145:in `block in render'
…3.0/gems/actionview-4.2.10/lib/action_view/template.rb: 333:in `instrument'

app/models/exercise_collection.rb
```diff
@@ -13,7 +13,12 @@ def exercise_working_times
   end
 
   def average_working_time
-    exercise_working_times.values.reduce(:+) / exercises.size
+    if exercises.empty?
+      0
+    else
+      values = exercise_working_times.values.reject { |v| v.nil?}
+      values.reduce(:+) / exercises.size
```
rteusner: still an error here, if values is empty...

PR #169, opened 2018-05-07
Edge compat

Parnswir:
IE11/Edge does not allow the trailing slash in the `protocol` property of a `URL` object, resulting in a silent fallback to `http` when opening up the websocket connection (in [execution.js.erb:7](https://github.com/openHPI/codeocean/blob/master/app/assets/javascripts/editor/execution.js.erb#L7)).

**Deployed configs should be adjusted as well** (even though the additional check in `editor/execution.js` technically makes this unnecessary).


rteusner:
lgtm

PR #175, opened 2018-06-05
IE url polyfill

Parnswir:
Basic run/submit flow should now work with IE 10 and 11. Same as in Edge, Internet Explorer raises an error when closing the websocket connection. We currently ignore this.


rteusner:
lgtm

PR #177, opened 2018-06-05
Parse runtime output on stop

Parnswir:
Close socket from container to server when killing the container in order to save run output


rteusner:
lgtm

PR #199, opened 2018-08-09
enable mnemosyne

rteusner:


config/mnemosyne.yml.example
```diff
@@ -0,0 +1,26 @@
...
+  # hostname:
+
+development:
+  <<: *common
+
+test:
+  enabled: false
+
+production:
+  <<: *common
```
MrSerth: What about adding a new line here?

config/mnemosyne.yml.example
```diff
@@ -0,0 +1,26 @@
+# Mnemosyne AMQP sink
```
MrSerth: What about copying the `mnemosyne.yml.example` for the Vagrant setup? If you do so, please ensure `mnemosyne.yml` is in `.gitignore`

config/mnemosyne.yml.example
```diff
@@ -0,0 +1,26 @@
...
+common: &common
+  # AMQP server endpoint
+  server: <%= ENV['XIKOLO_RABBITMQ_URL'] || 'amqp://localhost' %>
+  enabled: false
```
MrSerth: Is there a way to hide the `Submit trace ...` output in the console?

PR #200, opened 2018-08-09
Remove ab tests

rteusner:
Please check and shortly think about parts that I might have forgotten... ;-)


app/controllers/exercises_controller.rb
```diff
@@ -184,28 +184,8 @@ def implement
...
-        @hide_rfc_button = "true"
-      when :break_intervention_show_rfc
-        @show_break_interventions = showInterventions
-      when :no_intervention_show_rfc
-      when :rfc_intervention_show_rfc
-        @show_rfc_interventions = showInterventions
-    end
-
 
+    @show_rfc_interventions = (not user_solved_exercise and not user_got_enough_interventions) ? "true" : "false"
```
Parnswir: just call `to_s` on `(not user_solved_exercise and not user_got_enough_interventions)` 😃

app/models/proxy_exercise.rb
```diff
@@ -37,30 +37,14 @@ def get_matching_exercise(user)
...
-                Rails.logger.debug("find new matching exercise for user #{user.id}" )
-                begin
-                  find_matching_exercise(user)
-                rescue => e #fallback
-                  Rails.logger.error("finding matching exercise failed. Fall back to random exercise! Error: #{$!}" )
-                  @reason[:reason] = "fallback because of error"
-                  @reason[:error] = "#{$!}:\n\t#{e.backtrace.join("\n\t")}"
-                  exercises.where("expected_difficulty > 1").shuffle.first # difficulty should be > 1 to prevent dummy exercise from being chosen.
-                end
+              Rails.logger.debug("find new matching exercise for user #{user.id}" )
```
Parnswir: Does this actually work? (assigning matching_exercise) -> I think the log statement should move into the begin block. I fear otherwise you might be assigning whatever is returned by the logger to matching_exercise...

PR #204, opened 2018-08-22
Fix saving new exercise collections with exercises

Parnswir:
+ delete exercise collection items when exercises or collections are deleted


rteusner:
lgtm
will check on staging, then merge.

PR #205, opened 2018-08-23
Performance Optimizations:

MrSerth:
1. Add index on testrun.submissionId
2. replace lti_parameters.size with lti_parameters.present? to prevent COUNT query


rteusner:
lgtm

PR #207, opened 2018-09-11
Install latest updates and move all tests to spec folder

MrSerth:
My changes in the Gemfile should not change something except for removing unused gems that are no longer used. This also removes the Gemnasium badge from the Readme and fixes the open security updates.

We might want to test this branch on staging. However, when using Xikolo locally, everything (including LTI) worked as expected ;).


rteusner:
lgtm

PR #208, opened 2018-09-18
wip: Flowr

Parnswir:


app/controllers/flowr_controller.rb
```diff
@@ -0,0 +1,42 @@
+class FlowrController < ApplicationController
+
+  def insights
+    unless current_user
+      skip_authorization
```
MrSerth: What about using pundit for authorization management here? If we do so, we could remove lines 4-10.

app/policies/submission_policy.rb
```diff
@@ -8,7 +8,7 @@ def create?
     everyone
   end
 
-  [:download?, :download_file?, :render_file?, :run?, :score?, :show?, :statistics?, :stop?, :test?].each do |action|
+  [:download?, :download_file?, :render_file?, :run?, :score?, :show?, :statistics?, :stop?, :test?, :insights?].each do |action|
```
MrSerth: Which method are you referring here? I can't find `submission_controller#insights`.

MrSerth: Alright, I added a comment about that. Thanks!

MrSerth:
Except for the carets, I think we're done with this for now.

PR #211, opened 2018-09-30
Order files by name by default

franzliedke:
This makes the order of test files consistent and useful. I need this in the ruby course (starting tomorrow) to provide helpful feedback with each test file, which are meant to guide users to the complete solution, step by step.

Explicit ordering in the view (!) is now no longer necessary.

**Questions for reviewers:**
- Could this have (unwanted) side effects in other places where files (e.g. by users) are listed?


rteusner:
lgtm, there should be no side effects

PR #217, opened 2018-10-16
Update LOCAL_SETUP.md

jstaerk:
rails server will not be bridged to host unless listening on 0.0.0.0. There were no errors when deployed on vagrant/virtualbox on mac, nevertheless the password "admin" would still not work :-(

### DIFF
--- a/LOCAL_SETUP.md
+++ b/LOCAL_SETUP.md
@@ -59,7 +59,7 @@ All problems that have occurred resulted from a more restrictive rights manageme
 ### Start server
 vagrant ssh  
 cd /vagrant  
-rails s -p 3000  
+rails s -p 3000 -b 0.0.0.0
 
 ### Login to CodeOcean
 192.168.59.104:3000  


MrSerth:
Thanks for your contribution!

You're right, without specifying `-b 0.0.0.0` you can't access the Rails server from another host. In the default configuration, Rails will bind the server to the loopback interface on 127.0.0.1 / ::1 and thus prevent external requests. As you're running the server inside a virtual machine, you need to bind it to a non-loopback interface and specify 0.0.0.0 will bind it to all available interfaces. In the background, Vagrant will additionally ensure to redirect traffic to the desired port in the VM.

Regarding your login issue: It might help to re-seed the database. Try running these commands to reset your database (Warning: This will destroy all changes made in your database) inside the VM at your `/vagrant` path:

```
rake db:migrate:reset
rake db:seed
```

rteusner:
Thank you for the  feedback, I agree that it is likely to help others to know these settings.

PR #224, opened 2018-10-23
Fix retrieval of user_exercise_feedback and exercise

rteusner:


app/controllers/user_exercise_feedbacks_controller.rb
```diff
@@ -103,8 +103,8 @@ def to_s
   end
 
   def set_user_exercise_feedback
-    @exercise = Exercise.find(params[:user_exercise_feedback][:exercise_id])
-    @uef = UserExerciseFeedback.find_by(exercise_id: params[:user_exercise_feedback][:exercise_id], user: current_user)
+    @uef = UserExerciseFeedback.find(params[:id])
```
jgraichen: How about:

```suggestion
    @uef = UserExerciseFeedback.where(id: params[:id], user: current_user)
```

This limits the endpoint to only fetch owned feedbacks.

rteusner: yep, missed that.

PR #230, opened 2018-11-22
Specify `ExternalUser` as class for `user` in `RemoteEvaluationMapping`

MrSerth:
Otherwise, the download of submissions no longer works in Rails 5


rteusner:
good fix. lgtm

PR #231, opened 2018-11-22
Prevent creation of the Docker containers if just running a db rake task

MrSerth:
Otherwise, the rake task might fail if the database connection could not be established (e.g. when creating a new database).


rteusner:
lgtm

PR #232, opened 2018-11-22
Skip all docker images not tagged (otherwise, the method crashes)

MrSerth:


rteusner:
really do like it. :) 👍

PR #234, opened 2018-11-26
Migrate jQuery function use for ajax in case of success or failure

MrSerth:
Reenables users to reset their exercise implementation to the original version

The problem laid in old methods used from jQuery versions prior to 1.8 (which was released on August 9, 2012)

> Deprecation Notice from jQuery docs:
> 
> The jqXHR.success(), jqXHR.error(), and jqXHR.complete() callbacks will be deprecated in jQuery 1.8. To prepare your code for their eventual removal, use jqXHR.done(), jqXHR.fail(), and jqXHR.always() instead.

### DIFF
--- a/app/assets/javascripts/editor/editor.js.erb
+++ b/app/assets/javascripts/editor/editor.js.erb
@@ -436,7 +436,7 @@ configureEditors: function () {
       url: $('#editor').data('errors-url')
     });
     jqxhr.always(this.hideSpinner);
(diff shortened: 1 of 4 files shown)


ThomasStaubitz:
lgtm. Can you merge it or do I need to do that?

MrSerth:
Alright, I will do it in a few minutes as soon as Travis finished.

PR #227, opened 2018-11-15
Repair charts

rteusner:

PR #237, opened 2018-11-28
Rebase `repair-charts` onto `master`

MrSerth:
Updated version of #227


MrSerth:
I decided to show the x axis description only every second bar as it would be too much otherwise. We might change this as soon as an upper limit has been introduced.

Current status:

bildschirmfoto 2018-12-02 um 15 09 13

MrSerth:
``` js
return "<strong>Students:</strong><span style='color:orange'>" + d + "</span></br> <strong>Highest Value:</strong><span style='color:orange'>" +  maxVal + "</span></br> <strong> Average Value: </strong><span style='color:orange'>" + (sum/d).toFixed(2) +"</span>";
```

PR #238, opened 2018-11-29
Fix adding files to exercises (as admin)

MrSerth:
Prevent hidden selection field from being removed after cloning the dummy. Also ensure the caret looks nice and add the file extension if possible.


rteusner:
👍

PR #242, opened 2018-12-06
Add user (ID and type) to proxy exercise

MrSerth:
This is required for the existing policies to work


rteusner:
also adds public flag.

lgtm 👍

Issue #140, opened 2017-11-18
All exercises accessible by URL manipulation


flokX:
If you are signed in in the CodeOcean, you can switch between all excercises by changing the URL.

Ways to reproduce the bug:
1. Sign in to the CodeOcean by doing an exercise in openHPI
2. Now you can simply switch between different excercises by changing the excercise number: `https://codeocean.openhpi.de/exercises/[try random numbers]/implement`

tstaubitz:
Why do you consider this to be a bug? As far as we are concerned, it is ok to work on as many exercises as you want. You probably will run into a problem, however, if you try to submit the results of an exercises that you accessed by tampering with the URL to the LMS from which you started the initial exercise, if you're not using the platform standalone.

flokX:
In my opinion it's never good to get access to areas you don't need, but if it's a conscious decision I am going to close this PR.

franzliedke:
@tstaubitz This would actually help a lot to prevent some of the problems CodeOcean has had in regards to returning the right result for the wrong exercise.

Authentication via LTI should only ever give you access to exactly one exercise (i.e. the one identified by the provided token). If others weren't accessible, results could not be submitted incorrectly.

MrSerth:
I might have a look at this issue when checking the integration of multiple exercises on a worksheet but won't promise anything.

PR #244, opened 2018-12-11
Implement support for some basic embed options for work sheets via LTI

MrSerth:
I implemented the restrictions discussed a while ago but we could add more if needed (e.g. there is no time / run restriction yet).

The restrictions are present in the UI and enforced for server-side actions and thus should prevent unauthorized usage of the specified features. The restriction settings are put into the encrypted session cookie of the user and are applied for any request (independent from the policies which are not changed, so you might still open a different exercise by guessing the right URL, see #140).

As soon as the user signs out, the restrictions are removed and the session is terminated. As the cookie has a lifetime of `-1`, the session will automatically terminated once the browser quits. Closing the tab and reopen it won't invalidate the session cookie in most browsers. We might discuss whether to put the information somewhere else or actively delete the session.

This commit also fixes an issue with the flash messages being positioned too high and displayed for too long.


rteusner:
lgtm 👍

PR #246, opened 2018-12-12
Consider empty name and mail for external users

MrSerth:
These changes should prevent undesired access on empty names or empty email addresses. We might consider changing the pseudo_name to something cool, e.g. by using the [Faker](https://github.com/stympy/faker) gem or the already integrated [Forgery](https://github.com/sevenwire/forgery) gem.

We might also discuss why the display name has been introduced and in which situations not to use it (as I did for `external_user/show`).


rteusner:
lgtm 👍

PR #248, opened 2018-12-14
Improve groups with a view and always create a group

MrSerth:
As discussed for #245 

Teachers are now enabled to rename groups, however, we can't automatically get the course name as this resource is JWT-protected.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -105,7 +105,7 @@ en:
         external_id: External ID
         consumer: Consumer
         members: Members
-        member_count: Members Count
+        member_count: Member Count
       tag:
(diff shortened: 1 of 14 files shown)


config/locales/en.yml
```diff
@@ -100,6 +100,12 @@ en:
         files: Files
         score: Score
         user: Author
+      study_group:
+        name: Name
+        external_id: External ID
+        consumer: Consumer
+        members: Members
+        member_count: Members Count
```
MrSerth:
```suggestion
        member_count: Member Count
```

rteusner:
lgtm 👍

PR #251, opened 2019-01-29
Multiple Bugfixes

MrSerth:


app/controllers/application_controller.rb
```diff
@@ -20,7 +20,14 @@ def require_user!
...
   def render_not_authorized
     respond_to do |format|
-      format.html { redirect_to(request.referrer || :root, alert: t('application.not_authorized')) }
+      format.html do
+        if request.referrer.present? && request.referrer.include?(request.base_url)
+          destination = request.referrer
+        else
+          destination = :root
+        end
+        redirect_to(destination, alert: t('application.not_authorized'))
```
franzliedke: Rails 5 has [`redirect_back`](https://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_back), which should be just what you need here. (Be careful to include the `allow_other_host` parameter, though.)

MrSerth: Thanks, I adopted the code here!

rteusner:
lgtm 👍

PR #261, opened 2019-03-12
Add LA dashboard architecture

MrSerth:
This PR adds the general architecture required for the Learning Analytics dashboard. It also allows saving the study group id in the submission.

Unfortunately, I was required to slightly change the nginx config on staging to set the connection header to "Upgrade". It might be worth to check why the other current default is used, which always sets the connection header to `''`.

``` config
   location /cable {
     proxy_pass http://puma;
     proxy_http_version 1.1;
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     proxy_set_header Host $http_host;
     proxy_set_header Upgrade $http_upgrade;
     proxy_set_header Connection "Upgrade";
   }
```


app/assets/javascripts/channels/la_exercises.js
```diff
@@ -0,0 +1,250 @@
...
+            });
+        }
+
+        function handleWorkingTimeUpdate(data) {
+            const user_progress = data['user_progress'];
+            const additional_user_data = data['additional_user_data'];
+
+            const user = additional_user_data[additional_user_data.length - 1][0];
+            const position = userPosition[user.type + user.id]; // TODO validate: will result in undef. if not existent.
+            // TODO: Do update
```
rteusner: still open todos - postpone them for the meantime?

app/assets/javascripts/channels/la_exercises.js
```diff
@@ -0,0 +1,250 @@
...
+                    .attr("y", d => y(d[1]))
+                    .attr("height", d => y(d[0]) - y(d[1]))
+                    .transition()
+                    .attr("x", (d, i) => x(i))
+                    .attr("width", x.bandwidth());
+            }
+
+            $('#no_chart_data').addClass("d-none");
+            transitionStacked();
+            // transitionGrouped();
```
rteusner: is this a leftover?

app/assets/javascripts/channels/la_exercises.js
MrSerth: Yes, there are still open ToDos. For now, new working time data is sent to the frontend but not handled yet. I still need to incorporate merging the new data with the existing graph.

app/assets/javascripts/channels/la_exercises.js
MrSerth: Somehow. I'll remove it here and add a comment to the method explaining what it does (the method `transitionGrouped()` is never called but might be interesting in the future. Or, I just add a button to the top right corner of the graph that allows switching between the stacked view and the grouped view.

app/controllers/request_for_comments_controller.rb
```diff
@@ -149,7 +142,7 @@ def set_request_for_comment
 
     # Never trust parameters from the scary internet, only allow the white list through.
     def request_for_comment_params
-      # we are using the current_user.id here, since internal users are not able to create comments. The external_user.id is a primary key and does not require the consumer_id to be unique.
```
rteusner: thanks for cleaning this up

app/models/exercise.rb
```diff
@@ -90,6 +90,145 @@ def user_working_time_query
...
+      user_progress.each do |timings_array|
+        if timings_array.present? && timings_array.length != buckets + 1
+          timings_array[buckets] = nil
+        end
+      end
+    end
+
+    {user_progress: user_progress, additional_user_data: additional_user_data}
+  end
+
```
rteusner: I haven't checked these in my head. Lets step through them together later on.

PR #300, opened 2019-05-09
Fix working time calculation

MrSerth:
As RubyTree 1.0.1 was yanked, a bundle update is included here as well, updating the Ruby dependencies.


rteusner:
lgtm

PR #760, opened 2020-10-26
Store Linter Run Assessment

MrSerth:
Specifically optimized for PyLint

Bildschirmfoto 2020-10-26 um 21 18 05
Bildschirmfoto 2020-10-26 um 21 18 29

### DIFF
--- a/app/controllers/concerns/submission_scoring.rb
+++ b/app/controllers/concerns/submission_scoring.rb
@@ -12,14 +12,14 @@ module SubmissionScoring
         output = execute_test_file(file, submission)
         assessment = assessor.assess(output)
         passed = ((assessment[:passed] == assessment[:count]) and (assessment[:score]).positive?)
-        testrun_output = passed ? nil : 'message: ' + output[:message].to_s + "\n stdout: " + output[:stdout].to_s + "\n stderr: " + output[:stderr].to_s
+        testrun_output = passed ? nil : 'status: ' + output[:status].to_s + "\n stdout: " + output[:stdout].to_s + "\n stderr: " + output[:stderr].to_s
         unless testrun_output.blank?
           submission.exercise.execution_environment.error_templates.each do |template|
             pattern = Regexp.new(template.signature).freeze
             StructuredError.create_from_template(template, testrun_output, submission) if pattern.match(testrun_output)
           end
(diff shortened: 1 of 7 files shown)


app/controllers/concerns/submission_scoring.rb
```diff
@@ -12,22 +12,25 @@ def collect_test_results(submission)
         output = execute_test_file(file, submission)
         assessment = assessor.assess(output)
         passed = ((assessment[:passed] == assessment[:count]) and (assessment[:score]).positive?)
-        testrun_output = passed ? nil : 'message: ' + output[:message].to_s + "\n stdout: " + output[:stdout].to_s + "\n stderr: " + output[:stderr].to_s
+        testrun_output = passed ? nil : 'status: ' + output[:status].to_s + "\n stdout: " + output[:stdout].to_s + "\n stderr: " + output[:stderr].to_s
```
MrSerth, line 15: As discussed today, `message` is always `nil`.

app/controllers/concerns/submission_scoring.rb
```diff
@@ -12,22 +12,25 @@ def collect_test_results(submission)
...
         unless testrun_output.blank?
           submission.exercise.execution_environment.error_templates.each do |template|
             pattern = Regexp.new(template.signature).freeze
             StructuredError.create_from_template(template, testrun_output, submission) if pattern.match(testrun_output)
           end
         end
-        Testrun.new(
+        testrun = Testrun.create(
```
MrSerth, line 22: Any specific reason you decided for `.new().save` instead of `.save`? With `save`, the result is only `true` or `false` whether the entry was saved or not. With `create`, the newly created testrun is returned (and can be referenced later).

MrSerth:
Already merged and deployed, but happy to receive feedback!

app/controllers/concerns/submission_scoring.rb
rteusner: if this is the case, just omit / remove it.

app/controllers/concerns/submission_scoring.rb
rteusner: Nope, no specific reason here.

PR #767, opened 2020-10-29
Fix codeharbor-export policies

### LINKED ISSUE #765: Check CodeHarbor link before performing action
Users without a link to CodeHarbor are able to start the `export_external_check`. We should prevent that.

___

Sentry Issue: CODEOCEAN-9

```
NoMethodError: undefined method `check_uuid_url' for nil:NilClass
  app/services/exercise_service/check_external.rb:34:in `connection'
    Faraday.new(url: @codeharbor_link.check_uuid_url) do |faraday|
  app/services/exercise_service/check_external.rb:11:in `execute'
    response = connection.post do |req|
  app/services/service_base.rb:5:in `call'
    new(*args).execute
  app/controllers/exercises_controller.rb:125:in `export_external_check'
    codeharbor_check = ExerciseService::CheckExternal.call(uuid: @exercise.uuid, codeharbor_link: current_user.codeharbor_link)
...
(90 additional frame(s) were not displayed)
```

kkoehn:
fixes #765 

Added a check to determine if a user actually has the codeharbor_link configured before offering the export.

### DIFF
--- a/app/policies/exercise_policy.rb
+++ b/app/policies/exercise_policy.rb
@@ -19,8 +19,12 @@ class ExercisePolicy < AdminOrAuthorPolicy
     admin?
   end
 
-  [:clone?, :destroy?, :edit?, :update?, :export_external_check?, :export_external_confirm?].each do |action|
+  [:clone?, :destroy?, :edit?, :update?].each do |action|
     define_method(action) { admin? || teacher_in_study_group? || author? }
+  end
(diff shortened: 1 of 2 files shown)


MrSerth (approved):
I like having the checks done via the Policy! Don't know whether I would have been smart enough to implement it this way. Thanks!

PR #798, opened 2020-11-26
Add Rails admin

kkoehn:
Adds https://github.com/sferik/rails_admin to code_ocean
- added gem
- used user.admin? as authorization
- added entry to navigation bar
- Had to include a monkey patch to solve compatibility issues between kaminari (included in rails_admin) and will_paginate


Gemfile.lock
```diff
@@ -157,8 +164,26 @@ GEM
       oauth (>= 0.4.5, < 0.6)
     jbuilder (2.10.1)
       activesupport (>= 5.0.0)
+    jquery-rails (4.4.0)
+      rails-dom-testing (>= 1, < 3)
+      railties (>= 4.2.0)
+      thor (>= 0.14, < 2.0)
+    jquery-ui-rails (6.0.1)
+      railties (>= 3.2.16)
```
MrSerth, line 172: We use webpack to include these dependencies. However, I was unable to find a way to use these existing dependencies. This seems to be an open issue in RailsAdmin.

MrSerth (commented):
Thanks for investigating! I have two minor suggestions and an open question but would merge anyway.

MrSerth (approved):

MrSerth:
I deployed the changes and they work technically. However, loading RailsAdmin is super, super slow, and might take multiple minutes to render the dashboard. I would assume we need to disable some of the statistics on the dashboard in order to reduce the loading times.

Here is [some useful documentation](https://github.com/sferik/rails_admin/wiki/Dashboard-action).

PR #800, opened 2020-11-28
disable rails_admin dashboard

kkoehn:
The Dashboard is suspected to have performance problems.
Disable it.

@MrSerth are you able to see what kind of queries cause the slowdown on production? I also remember reading something about n+1 queries during the authorization checks and I can imagine that there are quite a few users in CO

### DIFF
--- a/config/initializers/rails_admin.rb
+++ b/config/initializers/rails_admin.rb
@@ -30,8 +30,11 @@ RailsAdmin.config do |config|
   # config.show_gravatar = true
 
   config.actions do
-    dashboard                     # mandatory
-    index                         # mandatory
+    # mandatory
+    dashboard do
+      statistics false
+    end
+    index # mandatory
     new
     export
     bulk_delete
@@ -50,10 +53,21 @@ RailsAdmin.config do |config|
     module WillPaginate
       module ActiveRecord
         module RelationMethods
-          def per(value = nil) per_page(value) end
-          def total_count() count end
-          def first_page?() self == first end
-          def last_page?() self == last end
+          def per(value = nil)
+            per_page(value)
+          end
+
+          def total_count()
+            count
+          end
+
+          def first_page?()
+            self == first
+          end
+
+          def last_page?()
+            self == last
+          end
         end
       end
       module CollectionMethods


kkoehn:
I just checked again: the pundit extension for rails_admin has a memory leak problem and it queries all users to check for the permissions, but we are NOT using it right now. I checked locally and the permission check works exactly as expected (Query wise)

MrSerth:
These are the most time-consuming queries according to rack-mini-profiler:

```
   (1439.3ms)  SELECT COUNT(*) FROM "events"

   (143983.1ms)  SELECT COUNT(*) FROM "files"

  CodeOcean::File Load (155905.1ms)  SELECT  "files".* FROM "files" ORDER BY "files"."name" DESC LIMIT $1  [["LIMIT", 1]]

   (2393.0ms)  SELECT COUNT(*) FROM "lti_parameters"

   (391.6ms)  SELECT COUNT(*) FROM "structured_errors"

   (1531.2ms)  SELECT COUNT(*) FROM "structured_error_attributes"

   (1702.9ms)  SELECT COUNT(*) FROM "submissions"

   (40784.9ms)  SELECT COUNT(*) FROM "testruns"

   (126.3ms)  SELECT COUNT(*) FROM "user_exercise_interventions"
```

Loading the page took more than five minutes in total:

```
Completed 200 OK in 350480ms (Views: 809.0ms | ActiveRecord: 348829.6ms)
```

There seems to be a problem with retrieving some information about files. As these are ordered by name (which is expensive and not required), the "Last created" column states "over 5 years ago". This is not correct.

kkoehn:
I suggest trying to disable the dashboard first and see what problem pops up next.
I've found an issue which mentions performance issues for pagination as well (since it also uses count(*)) we can refer to [1] if get encounter that problem too

[1] rails_admin#2699

MrSerth:
Okay, let's give it a try 👍

MrSerth:
I spent some time on rails_admin today. Unfortunately, I got a segmentation fault on a regular basis when building the assets. Hence, I disabled concurrent asset-building which resolved the issue for now (with https://github.com/openHPI/codeocean/commit/a5ba91394d628c2041baab37e4a36d8ba9c82fac).

MrSerth:
The page is much better now. However, when listing files (`CodeOcean::File`) it is still slow due to the `COUNT(*)` that is involved.

PR #879, opened 2021-02-16
Introduce new option for CodeHarbor

MrSerth:
This PR adds a config option to disable and enable CodeHarbor links platform-wide. If disabled, any access to views is prevented with Pundit and will raise a `NotAuthorizedException`.

⚠️  Please ensure to adapt your `code_ocean.yml` config accordingly. The new `[:codeharbor][:enabled]` setting is required. 

This PR also fixes the tests to ensure they work independently of the config option and improves some translations for CodeHarbor links.


MrSerth:
The RfC spec is still failing, ignoring for now...

PR #924, opened 2021-04-29
Improve LTI dev environment

MrSerth:
## What does it change?

* Adds a new default consumer in dev environment for hassle-free setup with Xikolo
* Change the default port in dev environment to 3333 (already using that for a long time).

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -11,8 +11,8 @@ Vagrant.configure(2) do |config|
   end
   config.vm.network 'forwarded_port',
                     host_ip: ENV['LISTEN_ADDRESS'] || '127.0.0.1',
-                    host: 3000,
-                    guest: 3000
+                    host: 7000,
+                    guest: 7000
   config.vm.synced_folder '.', '/home/vagrant/codeocean'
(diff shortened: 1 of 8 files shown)


Vagrantfile
```diff
@@ -11,8 +11,8 @@ Vagrant.configure(2) do |config|
   end
   config.vm.network 'forwarded_port',
                     host_ip: ENV['LISTEN_ADDRESS'] || '127.0.0.1',
-                    host: 3000,
-                    guest: 3000
+                    host: 3333,
+                    guest: 3333
```
franzliedke, line 15: Could you use something other than `3333` here, please? 🙏🏼 

I used to use that in teambuilder as well, but recently changed it to avoid a conflict with our video-player dev environment.

(How is it that there are so many port numbers, and people always choose the same ones? 😆 )

franzliedke (changes requested):

Vagrantfile
MrSerth: I updated the MR based on our discussion and will carry my beloved, three-year-old port to the grave 🪦

franzliedke (approved):

Vagrantfile
franzliedke: *pat on the back*

PR #930, opened 2021-05-14
Rebrush action sidebar for implement view

MrSerth:
This MR:
* Removes the "Files" folder from the File Tree
* Moves all buttons from the action sidebar to some other places, mostly the File Tree
  * If the file tree is hidden, the download button is shown below the editor (bottom-left)
  * Reset this file is not in the bottom-right corner below the editor
* Increases the transition speed when collapsing or expanding the sidebars
* Reduces the weight of the collapse and expand buttons (left and right sidebar)

Our goal was to make some space for the exercise description which might be moved to the left sidebar.

Old view:
Bildschirmfoto 2021-05-14 um 18 13 55

New view:
Bildschirmfoto 2021-05-14 um 18 13 17

### DIFF
--- a/app/assets/stylesheets/editor.css.scss
+++ b/app/assets/stylesheets/editor.css.scss
@@ -85,10 +85,10 @@ button i.fa-spin {
   display: none;
 }
 
-#autosave-label{
+#statusbar{
   visibility: hidden;
+  margin-top: .2em;
   height: 1.6em;
-  text-align:  right;
   color: #777;
   font-size: 0.8em;
 }
--- a/app/assets/stylesheets/editor.css.scss
+++ b/app/assets/stylesheets/editor.css.scss
@@ -102,8 +102,8 @@ button i.fa-spin {
 }
 
 .sidebar-col-collapsed {
-  -webkit-transition: width 2s;
-  transition: width 2s;
+  -webkit-transition: width .2s;
+  transition: width .2s;
   width:67px;
   float:left;
   min-height: 1px;
@@ -112,8 +112,8 @@ button i.fa-spin {
 }
 
 .sidebar-col {
-  -webkit-transition: width 2s;
-  transition: width 2s;
+  -webkit-transition: width .2s;
+  transition: width .2s;
   width:20%;
   float:left;
   min-height: 1px;
@@ -129,8 +129,8 @@ button i.fa-spin {
 }
 
 .output-col {
-  -webkit-transition: width 2s;
-  transition: width 2s;
+  -webkit-transition: width .2s;
+  transition: width .2s;
   width:40%;
   float:right;
   min-height: 1px;
@@ -141,8 +141,8 @@ button i.fa-spin {
 }
 
 .output-col-collapsed {
-  -webkit-transition: width 2s;
-  transition: width 2s;
+  -webkit-transition: width .2s;
+  transition: width .2s;
   width:67px;
   float:right;
   min-height: 1px;
@@ -173,16 +173,12 @@ button i.fa-spin {
 }
 
 .description-card-collapsed {
-  -webkit-transition: width 2s;
-  transition: width 2s;
   height: 0px;
   visibility: hidden;
 }
 
 .description-card {
   height: auto;
-  -webkit-transition: height 2s;
-  transition: height 2s;
   visibility: visible;
 }
 
--- a/app/views/exercises/_editor.html.slim
+++ b/app/views/exercises/_editor.html.slim
@@ -24,13 +24,29 @@
       = render('editor_button', data: {:'data-placement' => 'top', :'data-toggle' => 'tooltip', :'data-container' => 'body'}, icon: 'fa fa-trophy', id: 'assess', label: t('exercises.editor.score'), title: t('shared.tooltips.shortcut', shortcut: 'ALT + s')) unless @embed_options[:disable_score]
       - unless hide_rfc_button
         = render('editor_button', icon: 'fa fa-comment', id: 'requestComments', label: t('exercises.editor.requestComments'), title: t('exercises.editor.requestCommentsTooltip'))
+
     - @files.each do |file|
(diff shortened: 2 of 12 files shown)


app/assets/stylesheets/editor.css.scss
```diff
@@ -173,16 +173,14 @@ button i.fa-spin {
...
   height: 0px;
   visibility: hidden;
 }
 
 .description-card {
   height: auto;
-  -webkit-transition: height 2s;
-  transition: height 2s;
+  max-height: 300px;
+  overflow-y: auto;
```
franzliedke, line 183: IIRC, these new lines will be part 2. 😉

franzliedke (changes requested):

app/assets/stylesheets/editor.css.scss
MrSerth: Ups 🤭 You're right.

Dome-GER:
From what I know about the domain and got from the description, the changes look reasonable to me.

What I noticed: The download button (icon) appearance has changed. I liked the old icon better (instead of the simple down arrow).

MrSerth:
> From what I know about the domain and got from the description, the changes look reasonable to me.

Thanks for your feedback!

> What I noticed: The download button (icon) appearance has changed. I liked the old icon better (instead of the simple down arrow).

Yes, we exchanged that in favor of the visual weight the icon has. Here is a comparison of both icons:

Bildschirmfoto 2021-05-28 um 13 18 53
Bildschirmfoto 2021-05-28 um 13 21 16

spec/lib/file_tree_spec.rb
```diff
@@ -179,8 +179,22 @@
...
+      end
+    end
+
+    context 'with files' do
+      let(:files) { FactoryBot.build_list(:file, 10, context: nil, path: 'foo/bar/baz') }
+      let(:file_tree) { described_class.new(files) }
+      let(:js_tree) { file_tree.to_js_tree }
+
+      it 'produces the required JSON format with a file' do
+        # We ignore the root node and hence remove it here
```
franzliedke, line 194: I don't quite see any removal here.

That said, several tests in this file are very implementation-specific, where I would expect them to only test the behavior of `FileTree` that's observable from the outside.

(Not exactly a new problem, so not a blocker, but I wanted to bring it up.)

franzliedke (approved):
Sorry for taking so long. I like it!

Inline, two comments with general recommendations for this part of the codebase. Happy to pair on the tests, if you're interested.

spec/lib/file_tree_spec.rb
MrSerth: Both points are valid. I changed the comment to reflect that nothing is removed (in my assumption and compared to the old approach, the root node is _ignored_ and not considered for the comparison).

Regarding the test cases in this file: Indeed, many of the tests are implement-specific and not very general. As long as they work (and the implementation of the `FileTree` is not changed, I didn't consider changing it. Whenever it breaks, I would rather adopt them instead of tediously fixing them.

app/views/exercises/_editor.html.slim
```diff
@@ -24,13 +24,29 @@
...
       - file.read_only = true if @embed_options[:read_only]
       = render('editor_frame', exercise: exercise, file: file)
-    #autosave-label
-      = t('exercises.editor.lastsaved')
-      span
-      button style="display:none" id="autosave"
+
+    #statusbar.d-flex.justify-content-between
+      div
+        - if !@embed_options[:disable_download] && @exercise.hide_file_tree?
```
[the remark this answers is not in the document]
MrSerth: I couldn't agree more and totally see your points!

Because refactorings are not so easy, these views don't get touched too often (even though they would deserve it). The general code quality of the project is improvable and I am aware of many places I would like to refactor; especially within the `/exercise` views, controllers, and model(s).

As there is not any presenter in the app yet, I will check out some details and make a proposal regarding the integration and start using them. For now, I'll add it to my personal to-do list.

MrSerth:
> Sorry for taking so long. I like it!

Thanks for your review, I am glad about the progress we made so far ;)

> Inline, two comments with general recommendations for this part of the codebase. Happy to pair on the tests, if you're interested.

Of course, I am! We can pair on one of the existing tests or on new tests (many scenarios are still available for implementation 🙂)

spec/lib/file_tree_spec.rb
franzliedke: Thanks.
When the implementation doesn't change, unit tests are basically no-ops anyways. 🤣 

No rush, and certainly no blocker, it just stood out to me.

app/views/exercises/_editor.html.slim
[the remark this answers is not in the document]
franzliedke: > Because refactorings are not so easy, these views don't get touched too often (even though they would deserve it).

That's the crucial point IMO. And this is why the presenters and other refactorings would be the most valuable in this part of the codebase.

Have fun whenever you get around to it! Happy to help out if I can. I love this tricky challenge of pulling strings in complex parts of the code without breaking things, to extract small, clean parts, and see it gradually improve over time.

franzliedke:
> We can pair on one of the existing tests 

I hope we find a time some day. 🤗 Not a blocker for this MR, though.

app/views/exercises/_editor.html.slim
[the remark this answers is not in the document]
MrSerth: > That's the crucial point IMO. And this is why the presenters and other refactorings would be the most valuable in this part of the codebase.

I know that and the nice feeling of a clean code base providing the same features. For this PR, I will resolve the conversation and postpone the refactoring to a new ticket.

MrSerth:
I am going to merge this PR now and deploy it to production. If there is anything left (e.g., regarding the icons or adding a drop-down for the reset [instead of providing a separate link for the current file]), we can do that in a separate PR.

PR #1079, opened 2021-11-04
Sync execution environments

MrSerth:
This MR:

* Allows editing execution environments even when they didn't exist yet in the runner management
* Implements the `environments` and `delete_environment` methods for all strategies
* Ensures that execution environments are in sync at the runner management after failed edits

### DIFF
--- a/app/models/execution_environment.rb
+++ b/app/models/execution_environment.rb
@@ -33,10 +33,11 @@ class ExecutionEnvironment < ApplicationRecord
   before_validation :clean_exposed_ports
   validates :exposed_ports, array: {numericality: {greater_than_or_equal_to: 0, less_than: 65_536, only_integer: true}}
 
-  def set_default_values
-    set_default_values_if_present(permitted_execution_time: 60, pool_size: 0)
-  end
-  private :set_default_values
+  after_destroy :delete_runner_environment
+  after_save :working_docker_image?, if: :validate_docker_image?
+
+  after_rollback :delete_runner_environment, on: :create
+  after_rollback :sync_runner_environment, on: %i[update destroy]
 
   def to_s
     name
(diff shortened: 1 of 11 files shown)


app/models/execution_environment.rb
```diff
@@ -70,21 +76,48 @@ def valid_test_setup?
...
   def working_docker_image?
-    runner = Runner.for(author, self)
-    output = runner.execute_command(VALIDATION_COMMAND)
-    errors.add(:docker_image, "error: #{output[:stderr]}") if output[:stderr].present?
-  rescue Runner::Error::NotAvailable => e
-    Rails.logger.info("The Docker image could not be verified: #{e}")
+    sync_runner_environment
+    retries = 0
+    begin
+      runner = Runner.for(author, self)
```
mpass99, line 88: Here we request a new runner id via the `before_validation` hook of the Runner. This fails if we want to create an execution environment with an empty prewarming pool size (as db:seed does).
Should we adjust the prewarming pool size on the CodeOcean side for the moment of testing or provide an extra route or flag for test behaviour on the Poseidon side?

mpass99 (changes requested):

app/models/execution_environment.rb
MrSerth: Mh, that's a good finding I forgot (I actually tested the way to increase the pool size from 0 to another number but not the other way round). What if we skip the validation in case the new pool size is 0? The image cannot be used when the pool is empty and the next edit (which will increase the pool size) will allow verification.

mpass99: That sounds great!

MrSerth: Okay, great. I did so with b8e8f50 and additionally tested seeding manually. As a result, I also created #1082

mpass99 (approved):
Very cool functionalities you added here!

PR #1086, opened 2021-11-15
Add Rake task to GDPR delete users

MrSerth:
This MR adds a rake task to delete users in compliance with GDPR. It removes all personally identifiable information (PII) from external users stored in CodeOcean.

Usage:

```script
rake gdpr:delete_users TXT=./delete_request.txt CONSUMER=openHPI
```

with `delete_request.txt`:
```text
15489bf6-c174-4920-8452-5b111e967b1b
6f1ed1e6-56c4-407e-ac88-7f81387e3f0e
b38c0c9e-f676-41dd-b7fc-ddb2fc33c8ff
5df7f9af-c54e-4811-9e86-8e80bbf82106
3d7d8bf9-63e0-42cb-b961-cbeafe8d056e
```

### DIFF
--- a/lib/tasks/gdpr_delete.rake
+++ b/lib/tasks/gdpr_delete.rake
@@ -1,0 +1,78 @@
+# frozen_string_literal: true
+
+namespace :gdpr do
+  desc 'Deletes user accounts in accordance with GDPR. ' \
+       'Requires a text file with external user IDs (one per line) and ' \
+       'the corresponding consumer for deletion. Pass those via ' \
+       'TXT=/path/to/file.txt and CONSUMER=Name.'
+  task delete_users: :environment do
+    text_file_location = ENV.fetch('TXT')
+    consumer_name = ENV.fetch('CONSUMER')
+
+    if text_file_location.blank? || consumer_name.blank?
+      warn '--------------------------------------'
+      warn 'GDPR deletion tool to remove personal user data. USAGE:'
+      warn '--------------------------------------'
+      warn 'rake gdpr:delete_users TXT=./accounts.txt CONSUMER=openHPI'
+      exit 1
+    end
+
+    consumer = Consumer.where('lower(name) = ?', consumer_name.downcase).first
+
+    if consumer.blank?
+      warn '--------------------------------------'
+      warn "ERROR: The consumer '#{consumer_name}' was not found but is mandatory."
+      warn '--------------------------------------'
+      exit 2
+    end
+
+    user_ids = begin
+      File.readlines(text_file_location).map(&:strip).compact_blank
+    rescue SystemCallError => e
+      warn '--------------------------------------'
+      warn "ERROR: The file '#{text_file_location}' could not be opened: #{e.message}"
+      warn '--------------------------------------'
+      exit 3
+    end
+
+    unknown_user_ids = []
+    errored_user_ids = []
+    users_deleted = 0
+
+    user_ids.each do |user_id|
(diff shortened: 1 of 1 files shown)


MrSerth:
According to @cwillems, the integration with this Rake task is not yet planned. We're still waiting for the LTI consumer to create the corresponding files before proceeding.

PeterGillis:
LGTM 👍

lib/tasks/gdpr_delete.rake
```diff
@@ -0,0 +1,68 @@
...
+    rescue Errno => e
+      warn '--------------------------------------'
+      warn "ERROR: The file '#{text_file_location}' could not be opened: #{e.message}"
+      warn '--------------------------------------'
+      exit 3
+    end
+
+    unknown_user_ids = []
+    users_deleted = 0
+    file.each_line do |row|
```
Dome-GER, line 40: This shortcut comes in pretty handy when reading a file with data per line: `.readlines.map(&:chomp)`.
You'll get an array of IDs that can be iterated over.

lib/tasks/gdpr_delete.rake
```diff
@@ -0,0 +1,68 @@
...
+      next if row.blank?
+
+      user = ExternalUser.find_by(external_id: row.strip, consumer: consumer)
+      if user.blank?
+        unknown_user_ids << row
+        next
+      end
+
+      user.update!(name: 'Deleted User', email: nil)
```
Dome-GER, line 49: This will make the task fail if you cannot update the user. Is this intended? From the other error handling, I assume we'd like to continue and only add the user ID to the deleted users if we could update.

Dome-GER (approved):
Reasonable implementation, two smaller suggestions that might help to tweak this task.

lib/tasks/gdpr_delete.rake
MrSerth: That's a nice suggestion as well. At first, I decided against this method, as it requires reading the complete file at once for the processing. Depending on the size, the other solution might be better for very large files. But since we don't expect _that_ large files, I think this aspect is not worth keeping. Therefore, I changed the implementation according to your suggestion.

Next, one could consider changing the way we interact with database objects. Performance-wise, the current implementation has many drawbacks: For each (valid) line in the text file, we have one `SELECT` and one `UPDATE` query. We could also change this to use query all user IDs at once (i.e., with an `ExternalUser.where()` clause) or even using `update_all`. I refrained from doing so, since I wanted to focus on error-free processing and catch all errors that occur. For the bulk-operation, I am currently not aware of a suitable mechanism to identify those user IDs that were not used to fetch a record.

lib/tasks/gdpr_delete.rake
MrSerth: Indeed, that's true: The task would fail. While I wouldn't expect that a user update cannot be performed, I agree that another error handling would be more suitable. Therefore, I changed it accordingly.

MrSerth:
> Reasonable implementation, two smaller suggestions that might help to tweak this task.

Thanks for having a look and providing feedback. 🤗  I integrated your feedback and rebased this branch onto the latest master.

Once we have the corresponding counterpart in the LTI consumer, I'll continue merging this PR.

MrSerth:
@Dome-GER We don't have the corresponding Xikolo counterpart yet. Still, I was wondering whether we could finish the review. 🙏 

My proposal: Let's finish the review together and merge this PR. Then, it's off my table.

PR #1102, opened 2021-12-01
Use PostgreSQL version in Vagrant script

### LINKED ISSUE #1101: Vagrant script fails after new postgresql version 14 became available
## Problem

When running the `vagrant up` script as suggested in the documentation, the build failed with an error message
```
default: connection to server at "localhost" (127.0.0.1), port 5432 failed: FATAL:  password authentication failed for user "postgres"
```

The resulting VM did not function.

## Expected behaviour

Build results in a working VM.

## Steps to reproduce

Will happen for new vagrant runs ever since postgres 14 was released and became available

## Root Cause

This is due to line 30 of `provision.vagrant.sh` is configured to set up passwordless logins for localhost in a folder `/etc/postgresql/$postgres_version/main/pg_hba.conf` with `$postgres_version` being set to 13.

## Possible fix outline

There is two ways to fix it, either set the version to 14 in line 6, or call for an explicit version ` install -y postgresql-client-$postgres_version postgresql-$postgres_version` in line 27, or (better) a combination of both.

MrSerth:
Fixes #1101

### DIFF
--- a/provision/provision.vagrant.sh
+++ b/provision/provision.vagrant.sh
@@ -2,9 +2,9 @@
 
 ######## VERSION INFORMATION ########
 
-postgres_version=13
+postgres_version=14
 node_version=14
-ruby_version=2.7.2
+ruby_version=2.7.5
 
 ########## INSTALL SCRIPT ###########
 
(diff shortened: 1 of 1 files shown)


MrSerth:
Great, thank you! I will merge these changes and am happy to continue investigating the other issue as well. I created a new PR to tackle that: #1103

PR #1242, opened 2022-04-24
Refactor testrun table

MrSerth:
This MR:
- Adds additional information (exit code and execution status) to the testrun table
- Proposes a new store for log output (with the intention to replace the output column in the test run table)
- Includes a migration script for all existing outputs.
- Save detailed TestrunMessages for new code executions
- Limit output saved in TestrunMessages
- Tested migration script on staging

Missing parts once the migration is finished (another PR):

- Remove duplicated saving of output
- Remove unused enums for TestrunMessages

Closes https://github.com/openHPI/poseidon/issues/100

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -9,6 +9,7 @@ class SubmissionsController < ApplicationController
 
   before_action :require_user!
   before_action :set_submission, only: %i[download download_file render_file run score show statistics test]
+  before_action :set_testrun, only: %i[run score test]
   before_action :set_files, only: %i[download show]
   before_action :set_files_and_specific_file, only: %i[download_file render_file run test]
   before_action :set_mime_type, only: %i[download_file render_file]
@@ -103,6 +104,7 @@ class SubmissionsController < ApplicationController
 
         case event[:cmd].to_sym
           when :client_kill
+            @testrun[:status] = :client_kill
             close_client_connection(client_socket)
             Rails.logger.debug('Client exited container.')
           when :result, :canvasevent, :exception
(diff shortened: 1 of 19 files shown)


MrSerth:
The migration script should be fine now (I hope). It has been tested with 20 mio. existing Testruns and succeeded. I'll give it another try on staging before deploying to production but think it is fine now (and ready for a first round of review 😁)

app/controllers/submissions_controller.rb
```diff
@@ -169,14 +177,16 @@ def run
     close_client_connection(client_socket)
     Rails.logger.debug { "Running a submission timed out: #{e.message}" }
     @output = "timeout: #{@output}"
+    @status = :timeout
     extract_durations(e)
   rescue Runner::Error => e
     client_socket.send_data JSON.dump({cmd: :status, status: :container_depleted})
     close_client_connection(client_socket)
     Rails.logger.debug { "Runner error while running a submission: #{e.message}" }
+    @status = :container_depleted
```
mpass99, line 186: Why is here no `@passed = false` like for the other `:container_depleted`? Is `@passed` a test-specific field and `@exit_code` a run-specific field (like it is used in `migrate_messages`)?

mpass99 (commented):
Wow, that must have been a lot of work :muscle: As far as my Ruby knowledge allows me to give an opinion here, I would say that everything looks good - structured, clean and logically plausible!

app/controllers/submissions_controller.rb
MrSerth: Yes, exactly. In the `status` column, we store general information about the execution: Did it succeed and exited normally (status code 0)? Or did it fail? And if it failed, why (`failed` = user's fault, `timeout` = we cancelled, `out_of_memory` = OOM killer kicked in, ...). 

Additionally, we store the `exit_code`. That might be somehow duplicated with the `status` column: 
- `status = :ok` -> `exit_code = 0`
- `status = :out_of_memory` -> `exit_code = 137`
- `status = :timeout` -> no exit code
- `status = :container_depleted` -> no exit code
- `status = :failed` -> `exit_code >= 1 && exit_code <= 255 && exit_code != 137`

Additionally, we have the `passed` column. That is solely used for test executions, not for normal run executions. In that column, we store whether the test passed (in which case we don't store the output) or whether the test failed (in which case we have the output). It can be inferred from the status column as well:

- `status = :ok` -> `passed = true`
- `status = :failed` -> `passed = false`

I also considered reducing the number of columns, but I would still argue that some information are still valuable. Most probably, the `passed` column could be removed as I would be able to reconstruct it with the `cause` and `status` columns. Maybe that's something for another migration ;)

MrSerth:
> Wow, that must have been a lot of work 💪 As far as my Ruby knowledge allows me to give an opinion here, I would say that everything looks good - structured, clean and logically plausible!

Thanks for having a look! I am currently checking the remaining manual tasks for the migration (while the migration itself is being tested on staging) and look forward to continuing with the last part soon.

PR #1303, opened 2022-07-20
#1300 option to limit exercise statistics

Janis4411:
Add option to limit exercise statistics to user-interactions #1300

When in the statistics view of a certain exercise you can now toggle all autosave submissions on and off. There is also a label indicating the status of the current state to make it more user friendly.

Issue #1300, opened 2022-07-18
Add option to limit exercise statistics to user-interactions


MrSerth:
As a _teacher_,
I want to _see only user-triggered actions in the exercise statistics_
in order to _have a better overview of the users' progress and reduce the rendering time for the page_.

<hr>

Implementation Details:

- Limit `/external_users/{user_id}/exercises/{exercise_id}/statistics` to hide the submissions with a cause of `auto-save` by default
- Add a button to reload the page and show all submissions (same behaviour as of today)
- I would propose to implement the desired functionality without (much) additional JavaScript and instead reuse the existing Slim template and Ruby code.
- Teachers should not notice any difference (they don't have the permission to view the `auto-save` submissions anyway

Janis4411:
Feature completed.

PR #1304, opened 2022-07-20
Refactor external_user_statistics to own controller action

MrSerth:
This PR:
- is based on [this review comment](https://github.com/openHPI/codeocean/pull/1303#pullrequestreview-1044832274)
- keeps the same behaviour implemented for #1300
- refactors the `ExerciseController#statistics` method to be separated into two parts (one of the general exercise statistics and one for a specific user who solved the given exercise)
- does not yet tackle the large number of instance variables passed to the view (most code is just moved)

### DIFF
--- a/app/controllers/exercises_controller.rb
+++ b/app/controllers/exercises_controller.rb
@@ -492,13 +492,13 @@ working_time_accumulated: working_time_accumulated})
     # Render statistics page for one specific external user
 
     if policy(@exercise).detailed_statistics?
-      @submissions = Submission.where(user: @external_user,
+      submissions = Submission.where(user: @external_user,
         exercise_id: @exercise.id).in_study_group_of(current_user).order('created_at')
       @show_autosaves = params[:show_autosaves] == 'true'
-      @submissions = @submissions.where.not(cause: 'autosave') unless @show_autosaves
+      submissions = submissions.where.not(cause: 'autosave') unless @show_autosaves
       interventions = UserExerciseIntervention.where('user_id = ?  AND exercise_id = ?', @external_user.id,
         @exercise.id)
-      @all_events = (@submissions + interventions).sort_by(&:created_at)
+      @all_events = (submissions + interventions).sort_by(&:created_at)
       @deltas = @all_events.map.with_index do |item, index|
         delta = item.created_at - @all_events[index - 1].created_at if index.positive?
         delta.nil? || (delta > StatisticsHelper::WORKING_TIME_DELTA_IN_SECONDS) ? 0 : delta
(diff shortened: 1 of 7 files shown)


Janis4411 (approved):
Good change to keep everything a little cleaner.
Looks very good to me @MrSerth!

app/controllers/exercises_controller.rb
```diff
@@ -466,57 +467,58 @@ def reload
...
+        exercise_id: @exercise.id).in_study_group_of(current_user).final
+      @submissions = []
+      %i[before_deadline within_grace_period after_late_deadline].each do |filter|
+        relevant_submission = final_submissions.send(filter).latest
+        @submissions.push relevant_submission if relevant_submission.present?
       end
-      render locals: {
-        user_statistics: user_statistics,
-      }
+      @all_events = @submissions
```
Dome-GER, line 518: What is the reason for this assignment? Is the variable in this (prior) if branch just named differently? Can we align it?

spec/controllers/exercises_controller_spec.rb
```diff
@@ -236,8 +236,8 @@
     expect_template(:statistics)
   end
 
-  describe 'GET #statistics for external users' do
-    let(:perform_request) { get :statistics, params: params }
+  describe 'GET #external_user_statistics for external users' do
```
Dome-GER, line 239: This is duplicating the aspect of "external users" now.

This would be sufficient for me:
```suggestion
  describe 'GET #external_user_statistics' do
```

Dome-GER (changes requested):

app/controllers/exercises_controller.rb
MrSerth: No, the variable is not _just_ named differently. Until now, we passed two instance variables to the view: `@submissions` consisting of all user submissions (and thus having files, ...) and `@all_events` consisting of the submissions plus all interventions that occurred for the learner. In the view, we then present a timeline with the submissions and interventions. Therefore, my proposal would be to eliminate the `@submissions` instance variable, and just pass `@all_events`. In the view, I can then separate those again with `@all_events.filter{|event| event.is_a? Submission}`. This is still not perfect (and I like the idea of integrating a presenter) but would reduce the number of variables passed to the view.

In the case we are currently looking at (a user not having the `policy(@exercise).detailed_statistics?`, we do not show the interventions at all (as a measurement to increase privacy). Therefore, only those are passed and the interventions don't need to be added. Compare that with the other line above:
https://github.com/openHPI/codeocean/blob/d28a3a0eaf6cffa030ad365a6716208af1e4e205/app/controllers/exercises_controller.rb#L501

spec/controllers/exercises_controller_spec.rb
MrSerth: That's true, I'll change that! Thanks! :)

app/controllers/exercises_controller.rb
MrSerth: I made a proposal with my last commit 85f05712, please have look 😊

Dome-GER: Both options are not perfect (more variables vs. logic in the view). Typically, such logic like `submissions = @all_events.filter{|event| event.is_a? Submission}` should reside in a presenter. I don't mind too much.

Dome-GER (approved):

app/controllers/exercises_controller.rb
MrSerth: > Both options are not perfect (more variables vs. logic in the view)

I absolutely agree. Adding some presenters to the code base is also a nice suggestion, which I will consider for further refactoring. 

As discussed earlier today, I'll keep the new behaviour (with less variables) and continue with the other refactoring later. 

Thanks again for having a look :)

PR #1306, opened 2022-07-22
Rfc authentication

### LINKED ISSUE #1301: Add authentication token to RfC mails
As a _learner_,
I want to _view a Request for Comment (RfC) even when not logged in_
in order to _read replies from my fellow learners and do not get an error message_.

<hr>

Implementation Details:

- Generate user- and RfC-specific authorisation tokens
- Persist tokens and limit their validity to one week
- Send emails with the proper authentication token included
- Allow users to open the RfC with a single click as long as the token is still valid
- Preferably, the token should work for both, so-called internal and external users. If that doesn't work, external users are more important to enable the feature.

Janis4411:
Regarding issue: Add authentication token to RfC mails #1301

Via token authentication it is now possible to view a Request for Comment even when not logged in.

Closes #1301

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -15,7 +15,11 @@ class ApplicationController < ActionController::Base
   rescue_from ActionController::InvalidAuthenticityToken, with: :render_csrf_error
 
   def current_user
-    @current_user ||= ExternalUser.find_by(id: session[:external_user_id]) || login_from_session || login_from_other_sources || nil
+    @current_user ||= ExternalUser.find_by(id: session[:external_user_id]) ||
+                      login_from_session ||
+                      login_from_other_sources ||
+                      login_from_authentication_token ||
+                      nil
   end
 
   def require_user!
@@ -30,6 +34,13 @@ class ApplicationController < ActionController::Base
       ::Mnemosyne::Instrumenter.current_trace.meta['csrf_token'] = session[:_csrf_token]
       ::Mnemosyne::Instrumenter.current_trace.meta['external_user_id'] = session[:external_user_id]
     end
+  end
+
+  def login_from_authentication_token
+    token = AuthenticationToken.find_by(shared_secret: params[:token])
+    return unless token
+
+    auto_login(token.user) if token.expire_at.future?
   end
 
   def set_sentry_context
--- a/app/models/authentication_token.rb
+++ b/app/models/authentication_token.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+require 'securerandom'
+
+class AuthenticationToken < ApplicationRecord
+  include Creation
+  def self.generate!(user)
+    create!(
+      shared_secret: SecureRandom.hex(32),
+      user: user,
+      expire_at: 7.days.from_now
+    )
+  end
+end
--- a/app/models/authentication_token.rb
+++ b/app/models/authentication_token.rb
@@ -4,6 +4,7 @@ require 'securerandom'
 
 class AuthenticationToken < ApplicationRecord
   include Creation
+
   def self.generate!(user)
     create!(
       shared_secret: SecureRandom.hex(32),
--- a/spec/factories/authentication_token.rb
+++ b/spec/factories/authentication_token.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :authentication_token, class: 'AuthenticationToken' do
+    created_by_external_user
+    shared_secret { SecureRandom.hex(32) }
+    expire_at { 7.days.from_now }
+
+    trait :invalid do
+      expire_at { 8.days.ago }
+    end
+  end
+end
--- a/spec/features/authentication_spec.rb
+++ b/spec/features/authentication_spec.rb
@@ -35,12 +35,13 @@ describe 'Authentication' do
 
     context 'with no authentication token' do
       let(:request_for_comment) { create(:rfc_with_comment, user: user) }
+      let(:rfc_path) { request_for_comment_url(request_for_comment) }
 
       it 'denies access to the request for comment' do
-        mail.deliver_now
-        visit(rfc_link)
+        visit(rfc_path)
+        expect(page).not_to have_current_path(rfc_path)
         expect(page).not_to have_content(request_for_comment.exercise.title)
-        expect(response).to redirect_to(root_path)
+        expect(page).to have_current_path(root_path)
         expect(page).to have_content(I18n.t('application.not_authorized'))
       end
     end
@@ -60,8 +61,7 @@ describe 'Authentication' do
         it 'allows access to the request for comment' do
           mail.deliver_now
           visit(rfc_link)
-          expect(current_url).to be(rfc_link)
-          expect(response).to have_http_status :ok
+          expect(page).to have_current_path(rfc_link)
           expect(page).to have_content(request_for_comment.exercise.title)
         end
       end
@@ -72,8 +72,9 @@ describe 'Authentication' do
         it 'denies access to the request for comment' do
           mail.deliver_now
           visit(rfc_link)
+          expect(page).not_to have_current_path(rfc_link)
           expect(page).not_to have_content(request_for_comment.exercise.title)
-          expect(response).to redirect_to(root_path)
+          expect(page).to have_current_path(root_path)
           expect(page).to have_content(I18n.t('application.not_authorized'))
         end
       end
--- a/spec/features/authentication_spec.rb
+++ b/spec/features/authentication_spec.rb
@@ -32,6 +32,52 @@ describe 'Authentication' do
         expect(page).to have_content(I18n.t('sessions.create.failure'))
       end
     end
+
+    context 'with no authentication token' do
+      let(:request_for_comment) { create(:rfc_with_comment, user: user) }
+
+      it 'denies access to the request for comment' do
+        mail.deliver_now
+        visit(rfc_link)
+        expect(page).not_to have_content(request_for_comment.exercise.title)
+        expect(response).to redirect_to(root_path)
+        expect(page).to have_content(I18n.t('application.not_authorized'))
+      end
+    end
+
+    context 'with an authentication token' do
+      let(:user) { create(:learner) }
+      let(:request_for_comment) { create(:rfc_with_comment, user: user) }
+      let(:commenting_user) { InternalUser.create(attributes_for(:teacher)) }
+      let(:mail) { UserMailer.got_new_comment(request_for_comment.comments.first, request_for_comment, commenting_user) }
+      let(:rfc_link) { request_for_comment_url(request_for_comment, token: token.shared_secret) }
+
+      before { allow(AuthenticationToken).to receive(:generate!).with(user).and_return(token).once }
+
+      context 'when the token is valid' do
+        let(:token) { create(:authentication_token, user: user) }
+
+        it 'allows access to the request for comment' do
+          mail.deliver_now
+          visit(rfc_link)
+          expect(current_url).to be(rfc_link)
+          expect(response).to have_http_status :ok
+          expect(page).to have_content(request_for_comment.exercise.title)
+        end
+      end
+
+      context 'with an expired authentication token' do
+        let(:token) { create(:authentication_token, :invalid, user: user) }
+
+        it 'denies access to the request for comment' do
+          mail.deliver_now
+          visit(rfc_link)
+          expect(page).not_to have_content(request_for_comment.exercise.title)
+          expect(response).to redirect_to(root_path)
+          expect(page).to have_content(I18n.t('application.not_authorized'))
+        end
+      end
+    end
   end
 
   context 'when signed in' do
--- a/spec/mailers/user_mailer_spec.rb
+++ b/spec/mailers/user_mailer_spec.rb
@@ -14,7 +14,7 @@ describe UserMailer do
     end
 
     it 'sets the correct sender' do
-      expect(mail.from).to include(CodeOcean::Application.config.action_mailer[:default_options][:from])
+      expect(mail.from).to include('codeocean@hpi.de')
     end
 
     it 'sets the correct subject' do
@@ -45,7 +45,7 @@ describe UserMailer do
     end
 
     it 'sets the correct sender' do
-      expect(mail.from).to include(CodeOcean::Application.config.action_mailer[:default_options][:from])
+      expect(mail.from).to include('codeocean@hpi.de')
     end
 
     it 'sets the correct subject' do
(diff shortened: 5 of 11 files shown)


app/controllers/application_controller.rb
```diff
@@ -44,6 +49,10 @@ def set_sentry_context
   end
   private :set_sentry_context
 
+  def render_authentication_error
+    render_error t('')
```
Dome-GER, line 53: Something is missing here.

app/controllers/request_for_comments_controller.rb
```diff
@@ -144,6 +144,10 @@ def set_request_for_comment
     @request_for_comment = RequestForComment.find(params[:id])
   end
 
+  def add_authentication_token_to_params
```
Dome-GER, line 147: I don't belong here.

app/models/authentication_token.rb
```diff
@@ -0,0 +1,14 @@
+require 'securerandom'
+
+class AuthenticationToken < ApplicationRecord
+
+include Creation
```
Dome-GER, line 5: Indentation!

spec/mailers/user_mailer_spec.rb
```diff
@@ -60,4 +60,46 @@
       expect(mail.body).to include(reset_password_internal_user_url(user, token: user.reset_password_token))
     end
   end
+
+  describe '#got_new_comment' do
+
```
Dome-GER, line 65: Remove this additional empty line, please.

Dome-GER (changes requested):
Ignored the schema. Some minor suggestions.

app/controllers/application_controller.rb
```diff
@@ -32,6 +32,11 @@ def mnemosyne_trace
     end
   end
 
+  def login_from_authentication_token
+    token = AuthenticationToken.find_by(shared_secret: params[:token])
+    auto_login(token.user) if token&.user && token&.expire_at&.future?
```
Dome-GER, line 37: The scenario with and expired is not tested, is it?

MrSerth (changes requested):
Nice feature and very generell implementation approach, this will be well-used by future authentication features as well. 

@Janis4411 thanks for taking the time, I am convinced our learners will appreciate this new feature and like it very much! 🔥

spec/mailers/user_mailer_spec.rb
```diff
@@ -60,4 +60,122 @@
...
+    it 'includes the correct URL' do
+      expect(mail.body).to include(request_for_comment_url(request_for_comment))
+    end
+
+    it 'includes the correct authentication token' do
+      expect(mail.body).to include(token.shared_secret)
+    end
+
+    it 'sets a new authentication token' do
```
MrSerth, line 91:
```suggestion
    it 'creates a new authentication token' do
```

spec/mailers/user_mailer_spec.rb
```diff
@@ -60,4 +60,122 @@
...
+      expect { mail }.to change(AuthenticationToken, :count).by(1)
+    end
+
+    it 'sets a non expired authentication token' do
```
MrSerth, line 95:
```suggestion
    it 'sets a non-expired authentication token' do
```

spec/mailers/user_mailer_spec.rb
```diff
@@ -60,4 +60,122 @@
...
```
MrSerth, line 135:
```suggestion
    it 'sets a non-expired authentication token' do
```

spec/mailers/user_mailer_spec.rb
```diff
@@ -60,4 +60,122 @@
...
+
+    it 'includes the correct URL' do
+      expect(mail.body).to include(request_for_comment_url(subscription.request_for_comment))
+    end
+
+    it 'includes the correct authentication token' do
+      expect(mail.body).to include(token.shared_secret)
+    end
+
+    it 'sets a new authentication token' do
```
MrSerth, line 131:
```suggestion
    it 'creates a new authentication token' do
```

MrSerth (changes requested):
We should add another (request) spec to check that the token could be used to sign in iff it is not expired.

spec/factories/authentication_token.rb
```diff
@@ -0,0 +1,15 @@
...
+  factory :valid_authentication_token, class: 'AuthenticationToken' do
+    created_by_external_user
+    shared_secret { SecureRandom.hex(32) }
+    expire_at { 7.days.from_now }
+  end
+
+  factory :invalid_authentication_token, class: 'AuthenticationToken' do
+    created_by_external_user
+    shared_secret { SecureRandom.hex(32) }
+    expire_at { 8.days.ago }
```
Dome-GER, line 13: As we are modifying this attribute only, I would have expected it to be a `authentication_token` factory with and `:invalid` trait.

MrSerth (approved):
Looks great, thanks for taking so much time to test the behaviour and include the small improvements along the way 👏

spec/features/authentication_spec.rb
```diff
@@ -32,6 +32,46 @@
         expect(page).to have_content(I18n.t('sessions.create.failure'))
       end
     end
+
+    context 'with no authentication token' do
+      let(:request_for_comment) { create(:rfc_with_comment, user: user) }
+
+      it 'denies access to the request for comment' do
+        visit(request_for_comment_path(request_for_comment))
+        expect(page).to have_content(I18n.t('application.not_authorized'))
```
Dome-GER, line 41: This very much replicates implementation. We should check for expected content.
Similar below.

spec/features/authentication_spec.rb
```diff
@@ -32,6 +32,46 @@
...
+        let(:token) { create(:valid_authentication_token) }
+
+        it 'allows access to the request for comment' do
+          mail.deliver_now
+          visit(rfc_link)
+          expect(page).to have_content(request_for_comment.exercise.title)
+        end
+      end
+
+      context 'with a expired authentication token' do
```
Dome-GER, line 64:
```suggestion
      context 'with an expired authentication token' do
```

:nerd

spec/mailers/user_mailer_spec.rb
```diff
@@ -60,4 +60,122 @@
...
+    it 'sets the correct subject' do
+      expect(mail.subject).to eq(I18n.t('mailers.user_mailer.got_new_comment.subject', commenting_user_displayname: commenting_user.displayname))
+    end
+
+    it 'sets the correct receiver' do
+      expect(mail.to).to include(request_for_comment.user.email)
+    end
+
+    it 'includes the correct URL' do
+      expect(mail.body).to include(request_for_comment_url(request_for_comment))
```
Dome-GER, line 84: What URL is expected here? Same argument as above. The authentication token is supposed to be part of this URL, isn't it? Currently, it could appear anywhere in the e-mail body. A regular expression can help here.

spec/mailers/user_mailer_spec.rb
```diff
@@ -60,4 +60,122 @@
...
+
+    it 'includes the correct authentication token' do
+      expect(mail.body).to include(token.shared_secret)
+    end
+
+    it 'creates a new authentication token' do
+      expect { mail }.to change(AuthenticationToken, :count).by(1)
+    end
+
+    it 'sets a non-expired authentication token' do
```
Dome-GER, line 95: What is this supposed to be testing? Especially the second part (lines 98-99) looks suspicious to me.

spec/mailers/user_mailer_spec.rb
```diff
@@ -60,4 +60,122 @@
...
+  describe '#got_new_comment_for_subscription' do
+    let(:user) { create(:learner) }
+    let(:token) { AuthenticationToken.find_by(user: user) }
+    let(:request_for_comment) { create(:rfc_with_comment, user: user) }
+    let(:subscription) { Subscription.create(request_for_comment: request_for_comment, user: user) }
+    let(:from_user) { InternalUser.create(attributes_for(:teacher)) }
+    let(:mail) { described_class.got_new_comment_for_subscription(request_for_comment.comments.first, subscription, from_user).deliver_now }
+
+    it 'sets the correct sender' do
+      expect(mail.from).to include(CodeOcean::Application.config.action_mailer[:default_options][:from])
```
Dome-GER, line 112: Let's use the actual value here?

Dome-GER (changes requested):
Almost there 😉

app/controllers/application_controller.rb
Dome-GER: Fine for me. As commented on the changes, the scenarios testing the `future?` look a bit suscpicious.
Other than that, I'm very happy with the result.

MrSerth: That's true, we also discussed the tests with the `future?` earlier today. As a consequence, the feature spec were added to test the actual behaviour for expired tokens.

spec/features/authentication_spec.rb
```diff
@@ -32,6 +32,47 @@
...
+
+      before { allow(AuthenticationToken).to receive(:generate!).with(user).and_return(token).once }
+
+      context 'when the token is valid' do
+        let(:token) { create(:authentication_token, user: user) }
+
+        it 'allows access to the request for comment' do
+          mail.deliver_now
+          visit(rfc_link)
+          expect(page).to have_content(request_for_comment.exercise.title)
```
MrSerth, line 60: We could check here for the URL as well:
```
expect(current_url).to be(rfc_link)
```

spec/features/authentication_spec.rb
```diff
@@ -32,6 +32,47 @@
         expect(page).to have_content(I18n.t('sessions.create.failure'))
       end
     end
+
+    context 'with no authentication token' do
+      let(:request_for_comment) { create(:rfc_with_comment, user: user) }
+
+      it 'denies access to the request for comment' do
+        visit(request_for_comment_path(request_for_comment))
+        expect(page).to have_content(I18n.t('application.not_authorized'))
```
MrSerth, line 41: We could check here for the URLand the RfC title as well:
```
expect(current_url).not_to be(rfc_link)
expect(page).not_to have_content(request_for_comment.exercise.title)
```

spec/mailers/user_mailer_spec.rb
```diff
@@ -60,4 +60,107 @@
...
+
+  describe '#send_thank_you_note' do
+    let(:user) { create(:learner) }
+    let(:token) { AuthenticationToken.find_by(user: user) }
+    let(:request_for_comments) { create(:rfc_with_comment, user: user) }
+    let(:receiver) { InternalUser.create(attributes_for(:teacher)) }
+    let(:mail) { described_class.send_thank_you_note(request_for_comments, receiver).deliver_now }
+
+    it 'sets the correct sender' do
+      expect(mail.from).to include(CodeOcean::Application.config.action_mailer[:default_options][:from])
```
MrSerth, line 141:
```suggestion
      expect(mail.from).to include('codeocean@hpi.de')
```

spec/mailers/user_mailer_spec.rb
```diff
@@ -60,4 +60,107 @@
...
+      expect(mail.body).to include(request_for_comment_url(request_for_comments, token: token.shared_secret))
+    end
+
+    it 'creates a new authentication token' do
+      expect { mail }.to change(AuthenticationToken, :count).by(1)
+    end
+
+    it 'sets a non-expired authentication token' do
+      mail
+      # 5 minutes tolerance are allowed as the mail and the token are not created instantly at the same time.
```
MrSerth, line 162:
```suggestion
      # A five minute tolerance is allowed to account for the time difference between `now` and the creation timestamp of the token.
```

spec/mailers/user_mailer_spec.rb
```diff
@@ -60,4 +60,107 @@
...
+      expect(mail.body).to include(request_for_comment_url(subscription.request_for_comment, token: token.shared_secret))
+    end
+
+    it 'creates a new authentication token' do
+      expect { mail }.to change(AuthenticationToken, :count).by(1)
+    end
+
+    it 'sets a non-expired authentication token' do
+      mail
+      # 5 minutes tolerance are allowed as the mail and the token are not created instantly at the same time.
```
MrSerth, line 128:
```suggestion
      # A five minute tolerance is allowed to account for the time difference between `now` and the creation timestamp of the token.
```

spec/mailers/user_mailer_spec.rb
```diff
@@ -60,4 +60,107 @@
...
+      expect(mail.body).to include(request_for_comment_url(request_for_comment, token: token.shared_secret))
+    end
+
+    it 'creates a new authentication token' do
+      expect { mail }.to change(AuthenticationToken, :count).by(1)
+    end
+
+    it 'sets a non-expired authentication token' do
+      mail
+      # 5 minutes tolerance are allowed as the mail and the token are not created instantly at the same time.
```
MrSerth, line 93:
```suggestion
      # A five minute tolerance is allowed to account for the time difference between `now` and the creation timestamp of the token.
```

MrSerth (commented):
We could further simplify the specs by using `shared_example` and `include_example` (something I consider to be optional here)

spec/features/authentication_spec.rb
MrSerth: Okay, I got it (using I18n vs a string representation). Besides considering your comment for a refactoring of all specs, I would vote to leave it this way for now (to ensure all occurrences are consistent).

spec/features/authentication_spec.rb
```diff
@@ -32,6 +32,47 @@
...
+      end
+
+      context 'with an expired authentication token' do
+        let(:token) { create(:authentication_token, :invalid, user: user) }
+
+        it 'denies access to the request for comment' do
+          mail.deliver_now
+          visit(rfc_link)
+          expect(page).not_to have_content(request_for_comment.exercise.title)
+          expect(current_url).not_to be(rfc_link)
```
MrSerth, line 71:
```suggestion
          expect(response).to redirect_to(root_path)
```

spec/features/authentication_spec.rb
Dome-GER: That's rather a case for 

```suggestion
          expect(response).to have_http_status :ok
          expect(page).to have_content(request_for_comment.exercise.title)
```

spec/features/authentication_spec.rb
MrSerth: Using a `response` in feature specs is neither supported nor recommended (as this should be done in a request spec). Therefore, we could not implement the spec as suggested here, but rather sticked with the existing method to verify the URL and the exercise title.

spec/features/authentication_spec.rb
MrSerth: As mentioned above, using a response in feature specs is neither supported nor recommended (as this should be done in a request spec). Therefore, we could not implement the spec as suggested here, but rather sticked with the existing method to verify the URL and the exercise title.

spec/mailers/user_mailer_spec.rb
MrSerth: We removed the two suspicious lines and improved the spec. It now checks that a token valid for seven days is created for authentication.

MrSerth (approved):

Dome-GER (approved):
Nice result, I like it! 👍

PR #1317, opened 2022-08-04
Safe Comment Params

### LINKED ISSUE #876: Changing the file of a comment should not be possible
Every `Comment` belongs to a file and therefore has a file id. `CommentsController#update` only filters out the `reauest_id` in `comment_params_without_request_id`. So it is currently possible to change the file a comment belongs to. This should not be possible because the text of a comment relates to the commented file and would probably not relate properly to another file.

## Acceptance Criteria

- [ ] The controller does not accept the `file_id` parameter for the update action.
- [ ] The behavior described in the former point is tested (with a request spec?)
- [ ] Possible support in the frontend for this operation is removed

Janis4411:
Regarding issue: Changing the file of a comment should not be possible #876

Added strong params to comments_controller to prevent users from editing attributes which are not intended to be edited, also created specs to test this behaviour.

Closes #876

### DIFF
--- a/spec/controllers/comments_controller_spec.rb
+++ b/spec/controllers/comments_controller_spec.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe CommentsController do
+  let(:user) { create(:learner) }
+  let(:rfc_with_comment) { create(:rfc_with_comment, user: user) }
+  let(:comment) { rfc_with_comment.comments.first }
+  let(:updated_comment) { comment.reload }
+  let(:perform_request) { proc { put :update, format: :json, params: {id: comment.id, comment: comment_params} } }
+
+  before do
+    allow(controller).to receive(:current_user).and_return(user)
+    perform_request.call
+  end
+
+  describe 'PUT #update' do
+    context 'with valid params' do
+      let(:comment_params) { {text: 'test100'} }
+
+      it 'saves the permitted changes' do
+        expect(updated_comment.text).to eq('test100')
+      end
+
+      expect_http_status(:ok)
+    end
+
+    context 'with additional params' do
+      let(:comment_params) { {text: 'test100', row: 5, file_id: 50} }
+
+      it 'applies the permitted changes' do
+        expect(updated_comment.row).not_to eq(5)
+        expect(updated_comment.file_id).not_to eq(50)
+        expect(updated_comment.row).to eq(1)
+        expect(updated_comment.file_id).to eq(comment.file_id)
+        expect(updated_comment.text).to eq('test100')
+      end
+
+      expect_http_status(:ok)
+    end
+  end
+end
--- a/spec/factories/request_for_comment.rb
+++ b/spec/factories/request_for_comment.rb
@@ -13,7 +13,7 @@ FactoryBot.define do
     factory :rfc_with_comment, class: 'RequestForComment' do
       after(:create) do |rfc|
         rfc.file = rfc.submission.files.first
-        Comment.create(file: rfc.file, user: rfc.user, text: "comment for rfc #{rfc.question}")
+        Comment.create(file: rfc.file, user: rfc.user, row: 1, text: "comment for rfc #{rfc.question}")
       end
     end
   end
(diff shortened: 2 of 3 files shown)


Janis4411 (commented):
Looks good me. :)
The spec file could test more controller actions than just the update method, but as this issue resolved only around the update method, this should be sufficient.

What do you guys think @MrSerth, @Dome-GER?

MrSerth (approved):
Thanks for the changing the update method and adding some awesome specs! This improves the architecture and will prevent malicious data manipulations in the future! :nerd_face::+1:

spec/controllers/comments_controller_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  before do
+    allow(controller).to receive(:current_user).and_return(user)
+    perform_request.call
+  end
+
+  describe 'PUT #update' do
+    context 'with valid params' do
+      let(:comment_params) { {text: 'test100'} }
+
+      it 'saves the permited changes' do
```
Dome-GER, line 21:
```suggestion
      it 'saves the permitted changes' do
```

🤓 Here and below.

Small improvement potential: Combine the scenarios (as the first one is included in the second). A single scenario (`it`) along `'applies the permitted changes'` should be sufficient.

spec/factories/request_for_comment.rb
```diff
@@ -13,7 +13,7 @@
     factory :rfc_with_comment, class: 'RequestForComment' do
       after(:create) do |rfc|
         rfc.file = rfc.submission.files.first
-        Comment.create(file: rfc.file, user: rfc.user, text: "comment for rfc #{rfc.question}")
+        Comment.create(file: rfc.file, user: rfc.user, row: 1, text: "comment for rfc #{rfc.question}")
```
Dome-GER, line 16: Could you please quickly explain what the `row` is used for? Does it indicate the row (as in line of code) a comment refers to?

spec/controllers/comments_controller_spec.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe CommentsController do
```
Dome-GER, line 5: Typically, you would prefer a request spec over the controller spec. But if I remember correctly, the project only has controller specs, right?

Dome-GER (changes requested):
Looks good, please don't consider my comments a blocker.

spec/factories/request_for_comment.rb
Janis4411: Exactly, it refers to the line of code where the comment was made.

spec/controllers/comments_controller_spec.rb
Janis4411: Exactly. It has no request specs so far, so we sticked to controller specs.

spec/controllers/comments_controller_spec.rb
Janis4411: Thanks for correcting the wrong spelling.
I also changed the spec to have one instead of two scenarios.

spec/controllers/comments_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      let(:comment_params) { {text: 'test100'} }
+
+      it 'saves the permitted changes' do
+        expect(updated_comment.text).to eq('test100')
+      end
+
+      expect_http_status(:ok)
+    end
+
+    context 'with invalid params' do
```
Janis4411, line 28: Wording might not be perfect here, but I'd say it's good enough.

```
context 'with invalid params' do
let(:comment_params) { {text: 'test100', row: 5, file_id: 50} }

      it 'applies the permitted changes' do
```

Janis4411 (commented):
Applied ideas of @Dome-GER regarding wording and cleaner specs.
@MrSerth are you happy with the changes? :smile:

spec/controllers/comments_controller_spec.rb
Dome-GER: It's rather `'with additional params'`.

Dome-GER (approved):

spec/controllers/comments_controller_spec.rb
MrSerth: Thanks, I made the adjustment and will proceed with merging once the pipeline passes.

MrSerth (approved):

PR #1320, opened 2022-08-08
Disable score button

### LINKED ISSUE #256: Disable / Hide score button if no tests are available
As a learner, I would like not to be confused with the option to score my code if there are no tests available. So far, some exercises have a string, such as the following, in the description to clarify this:

`Hint: This “exercise” does not yield any points, there are no tests for the code here.`

Janis4411:
Regarding issue: Disable / Hide score button if no tests are available #256

Added functionality that if an exercise has no tests, the score button is now disabled.

closes #256

### DIFF
--- a/app/controllers/exercises_controller.rb
+++ b/app/controllers/exercises_controller.rb
@@ -324,6 +324,8 @@ class ExercisesController < ApplicationController
       end
     end
 
+    @embed_options[:disable_score] = true unless @exercise.teacher_defined_assessment?
+
     @hide_rfc_button = @embed_options[:disable_rfc]
 
     @search = Search.new
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -83,7 +83,7 @@ class SubmissionsController < ApplicationController
       client_socket = tubesock
 
       client_socket.onopen do |_event|
-        kill_client_socket(client_socket) if @embed_options[:disable_run]
+        return kill_client_socket(client_socket) if @embed_options[:disable_run]
       end
 
       client_socket.onclose do |_event|
@@ -199,7 +199,7 @@ class SubmissionsController < ApplicationController
     hijack do |tubesock|
       tubesock.onopen do |_event|
         switch_locale do
-          kill_client_socket(tubesock) if @embed_options[:disable_score]
+          return kill_client_socket(tubesock) if @embed_options[:disable_score] || !@submission.exercise.teacher_defined_assessment?
 
           # The score is stored separately, we can forward it to the client immediately
           tubesock.send_data(JSON.dump(@submission.calculate_score))
@@ -226,7 +226,7 @@ class SubmissionsController < ApplicationController
     hijack do |tubesock|
       tubesock.onopen do |_event|
         switch_locale do
-          kill_client_socket(tubesock) if @embed_options[:disable_run]
+          return kill_client_socket(tubesock) if @embed_options[:disable_run]
 
           # The score is stored separately, we can forward it to the client immediately
           tubesock.send_data(JSON.dump(@submission.test(@file)))
--- a/app/models/exercise.rb
+++ b/app/models/exercise.rb
@@ -205,6 +205,10 @@ class Exercise < ApplicationRecord
     "
   end
 
+  def teacher_defined_assessment?
+    files.any?(&:teacher_defined_assessment?)
+  end
+
   def get_working_times_for_study_group(study_group_id, user = nil)
     user_progress = []
     additional_user_data = []
--- a/spec/features/editor_spec.rb
+++ b/spec/features/editor_spec.rb
@@ -23,6 +23,7 @@ describe 'Editor', js: true do
     }]
   end
   let(:user) { create(:teacher) }
+  let(:exercise_without_test) { create(:tdd) }
 
   before do
     visit(sign_in_path)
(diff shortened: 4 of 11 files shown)


app/controllers/exercises_controller.rb
```diff
@@ -37,6 +37,10 @@ def max_intervention_count_per_exercise
     1
   end
 
+  def teacher_defined_test?
```
Janis4411, line 40: I was not sure what would be cleaner here.

Either you could do the check within one "big" line:
`@embed_options[:disable_score] = true unless @exercise.files.find {|f| f.role == 'teacher_defined_test' }`

Or, as I did, you could define a method to check if an exercise has a test and then use this method.

MrSerth: I would rather add a model method to the `Exercise`. This could be reused in more cases. Here is a proposal:

```ruby
def teacher_defined_assessment?
  files.any?(&:teacher_defined_assessment?)
end
```

MrSerth (commented):
Looks good so far, we might just want to add some tests and then proceed with merging 🙂

Thanks a lot!

app/models/exercise.rb
```diff
@@ -205,6 +205,10 @@ def study_group_working_time_query(exercise_id, study_group_id, additional_filte
     "
   end
 
+  def teacher_defined_assessment?
+    files.any?(&:teacher_defined_assessment?)
+  end
```
Janis4411, line 210: I did not know, that there was already a method to check for `teacher_defined_tests`. 
So thanks @MrSerth for the much cleaner solution. :)

spec/features/editor_spec.rb
```diff
@@ -93,6 +94,22 @@
...
+  end
+
+  context 'when an exercise has no teacher_defined_tests' do
+    it 'disables the score button' do
+      visit(implement_exercise_path(exercise_without_test))
+      expect(page).to have_content(exercise_without_test.title)
+      expect(page).not_to have_css('#assess')
+    end
+  end
+
```
Janis4411, line 112: This is the cleanest solution I found, if you have any advices on making it cleaner, I would love to hear your thoughts.

MrSerth (commented):
Great, I like how this is going! 💪 

Below are a few comments for further improvement. I also thought of denying the score and submit actions for users, so that they cannot be triggered by manipulating the URL.

1. For the score action, we might extend the following line:
https://github.com/openHPI/codeocean/blob/9587ecb2b11205701067c0749809e64c895387fb/app/controllers/submissions_controller.rb#L207
2. For the submit action, I thought of changing the policy:
https://github.com/openHPI/codeocean/blob/2141fdb1ac196df6ccb85170c2d4713f453029c2/app/policies/exercise_policy.rb#L32-L34

The following changes might do the trick there:
```ruby
%i[implement? working_times? intervention? search? reload?].each do |action|
  define_method(action) { everyone }
end

def submit?
  everyone && @record.teacher_defined_assessment?
end
```

MrSerth:
@Janis4411 I took the liberty to finish this MR and integrate the feedback I added earlier. Please have a look whether everything is correct and can be merged. 

//cc @Dome-GER

app/controllers/submissions_controller.rb
```diff
@@ -87,7 +87,7 @@ def run
       client_socket = tubesock
 
       client_socket.onopen do |_event|
-        kill_client_socket(client_socket) if @embed_options[:disable_run]
+        return kill_client_socket(client_socket) if @embed_options[:disable_run]
```
MrSerth, line 90: Without returning, the remaining code is still execution. Therefore, these return statements have been added.

MrSerth (approved):

spec/policies/exercise_policy_spec.rb
```diff
@@ -122,6 +122,28 @@
     end
   end
 
+  %i[submit?].each do |action|
```
Dome-GER, line 125: This looks suspicious. Tests seem to be very specific to the action. Why do the `each` stuff then?
I didn't look into details.

Dome-GER (approved):

spec/policies/exercise_policy_spec.rb
MrSerth: Thanks for having a look. I simplified the test cases by removing the iteration with the `each` wherever possible. This has been done for all permission checks in a dedicated commit.

spec/features/editor_spec.rb
```diff
@@ -93,6 +94,22 @@
     end
   end
 
+  context 'when an exercise has one or more teacher-defined assessments' do
+    it 'displays the score button' do
+      visit(implement_exercise_path(exercise))
+      expect(page).to have_content(exercise.title)
+      expect(page).to have_css('#assess')
```
Dome-GER, line 101: We probably want to check for a more meaningful (semantic) attribute here. This is a bit brittle. But fine for now, I guess.

Dome-GER (approved):
Looks good. Thanks for taking care of the small addition.

spec/features/editor_spec.rb
MrSerth: That's true, what do you think if I swap it to the following? This would allow to check for the behaviour (having the test to trigger the action) instead of the specific implementation with the element ID.
```suggestion
      expect(page).to have_content(I18n.t('exercises.editor.score'))
```

Dome-GER: Better 👍

MrSerth: Thanks, I've done so 🙂

Janis4411 (commented):
Looks good to me Sebastian! :)

PR #1349, opened 2022-08-31
Change HTTP status codes for Poseidon

MrSerth:
See https://github.com/openHPI/poseidon/pull/159 for more details.

@mpass99 I would appreciate if you could check my changes and especially verify that I didn't miss anything -- thanks!


mpass99 (approved):
Everything looks great with this PR :+1:

I just had some small notes during my review:
- l. 37 f.: This route does not return 404
- l. 258: In the execute route Poseidon just parses the passed parameters and temporarily saves them. I do not see a reason to delete the runner here. If we pass the parameters in a wrong format that's bad but it is not influenced by a (wrong) state of the runner.
- connection.rb `on_error`: Here the codes 401, 404, 410 and 500 could appear. Should the codes be interpreted like in the `handle_error` or is the general unknown error fine?

MrSerth:
> Everything looks great with this PR 👍

Great, thanks for having a detailed look and further finding issues!

<hr>

> * l. 37 f.: This route does not return 404

Fixed by removing the handling.

> * l. 258: In the execute route Poseidon just parses the passed parameters and temporarily saves them. I do not see a 
reason to delete the runner here. If we pass the parameters in a wrong format that's bad but it is not influenced by a (wrong) state of the runner.

That's true, I removed that code path

> * connection.rb `on_error`: Here the codes 401, 404, 410 and 500 could appear. Should the codes be interpreted like in the `handle_error` or is the general unknown error fine?

I tried looking into this as well, good point. However, I am afraid we cannot improve the handling much without parsing the error message by hand (which doesn't sound right). When establishing a connection, the [websocekt-driver-ruby](https://github.com/faye/websocket-driver-ruby) gem just [checks for the HTTP status code of 101](https://github.com/faye/websocket-driver-ruby/blob/b10eb2bf3e92f92fc3b5319de29b37949927f5f2/lib/websocket/driver/client.rb#L98-L100) and only [emits the error in a string-representation](https://github.com/faye/websocket-driver-ruby/blob/b10eb2bf3e92f92fc3b5319de29b37949927f5f2/lib/websocket/driver/client.rb#L90) if it doesn't match 101.

mpass99:
Nice, ok :+1:

PR #1352, opened 2022-09-02
Refactor methods and according specs

Janis4411:
Refactored send_thank_you_note method and authentication_token method as well as the according specs.


Janis4411:
Awesome, I'm glad I could be of a little help! 😄

PR #1356, opened 2022-09-06
Add Role to Study Group Membership

Janis4411:
This PR:

- adds a new authorisation concept based on the study group memberships
- migrates all current data; external users will be need to start an exercise through LTI to gain teacher privileges 
- creates a default study group for each consumer
- allows internal users to get assigned to study groups
- includes many smaller refactoring where appropriate


app/models/user.rb
```diff
@@ -38,6 +38,18 @@ def external_user?
     is_a?(ExternalUser)
   end
 
+  def learner?
+    role == 0
```
Janis4411, line 42: rubocop proposes `role.zero?` instead of `role == 0`
But I think we should use the same style for all three functions, right?

app/models/study_group_membership.rb
```diff
@@ -5,4 +5,7 @@ class StudyGroupMembership < ApplicationRecord
   belongs_to :study_group
 
   validates :user_id, uniqueness: {scope: %i[user_type study_group_id]}
+  validates :role, inclusion: {in: %w[teacher learner]}
+
+  enum role: { leaner: 0, teacher: 1, admin: 2 }
```
Janis4411, line 10: I should remove trailing whitespaces in the hash.

db/migrate/20220906142041_add_admin_to_external_users.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddAdminToBothUserTypes < ActiveRecord::Migration[6.1]
```
Janis4411, line 3: Not sure the naming is correct here.

db/migrate/20220906142603_remove_role_from_external_users.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class RemoveRoleFromBothUserTypes < ActiveRecord::Migration[6.1]
```
Janis4411, line 3: Same as above.

PR #1473, opened 2022-12-13
fix file_type selection for Makefile

### LINKED ISSUE #1466: Cannot import ProFormA task with "Makefile"
A ProFormA task created in CodeHarbor with a Makefile cannot be imported (or rather displayed) correctly in CodeOcean.

I already identified the cause:

https://github.com/openHPI/codeocean/blob/574e99eddd284cba9df3075c5c32fa311c923964/app/services/proforma_service/convert_task_to_exercise.rb#L84-L101

Here, we get the `extname` of the file:

https://github.com/openHPI/codeocean/blob/574e99eddd284cba9df3075c5c32fa311c923964/app/services/proforma_service/convert_task_to_exercise.rb#L85

For a `Makefile`, however, this is empty:

```rb
File.extname("Makefile")
 => "" 
```

Therefore, the Makefile (which is also a file type in CodeOcean) is not attached.

Closes CODEOCEAN-G3
Closes CODEOCEAN-GW

kkoehn:
fixes #1466 

- [x] add filetype not nil validation for file

### DIFF
--- a/app/assets/javascripts/exercises.js.erb
+++ b/app/assets/javascripts/exercises.js.erb
@@ -264,7 +264,7 @@ $(document).on('turbolinks:load', function () {
     var inferFileAttributes = function () {
         $(document).on('change', 'input[type="file"]', function () {
             var filename = $(this).val().split(/\\|\//g).pop();
-            var file_extension = '.' + filename.split('.')[1];
+            var file_extension = filename.includes('.') ? '.' + filename.split('.')[1] : '';
             var file_type = findFileTypeByFileExtension(file_extension);
             var name = filename.split('.')[0];
             var parent = $(this).parents('li');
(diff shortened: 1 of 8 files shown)


MrSerth (changes requested):
I am sorry, but think this fix is not sufficient.

First, if the "extension" is blank or nil, no valid `FileType` is created: `Name can't be blank`. This happens because no valid name is supplied.

Additionally, I just noticed that we have two different situations in our data regarding the file extension:

On our staging system, we have a file extension of `nil`, where our production data has a file extension of `""`. Also, whenever a file type with a file extension of `nil` is updated through the UI, it will be updated to `""`. Should we further look into this and unify the file extension (e.g., with a database migration and / or some additional validations...?)

With the old version, it probably worked on production but broke on staging and with the current version, it probably works on staging but broke on production.

Should we also add a migration to prevent an empty file_type?

MrSerth (approved):
Thanks, looks good and works as expected!

PR #1536, opened 2023-02-10
Enable Sentry instrumentation for WebSocket connection

MrSerth:
As a link says more than 1000 words, have a look [here](https://codeocean.sentry.io/performance/codeocean:95ec2d16cb3248dca12d40bae039724f/?breakdown=none&environment=development&project=5667283&query=http.method%3AGET&showTransactions=p95&sort=-transaction.duration&statsPeriod=24h&transaction=SubmissionsController%23run) 😁 
An old trace without the additional instrumentation is listed [here](https://codeocean.sentry.io/performance/codeocean:8fca428ace36487ebe98917358239302/?environment=production&project=5667283&query=http.method%3AGET+tags%5Btransaction%5D%3ASubmissionsController%23run&referrer=performance-transaction-summary&statsPeriod=24h&transaction=SubmissionsController%23run&unselectedSeries=p100%28%29) for example.

The new part are:
1. Sending the Sentry headers for the WebSocket connection
2. enabling breadcrumbs for the WebSocket connection
3. enabling instrumentation for all parts running in an EventMachine reactor.

Related to https://github.com/openHPI/poseidon/issues/270

### DIFF
--- a/lib/runner/connection.rb
+++ b/lib/runner/connection.rb
@@ -8,6 +8,8 @@ class Runner::Connection
   EVENTS = %i[start exit stdout stderr].freeze
   WEBSOCKET_MESSAGE_TYPES = %i[start stdout stderr error timeout exit].freeze
   BACKEND_OUTPUT_SCHEMA = JSONSchemer.schema(JSON.parse(File.read('lib/runner/backend-output.schema.json')))
+  SENTRY_OP_NAME = 'websocket.client'
+  SENTRY_BREADCRUMB_CATEGORY = 'net.websocket'
 
   # @!attribute start_callback
   # @!attribute exit_callback
@@ -18,7 +20,12 @@ class Runner::Connection
 
   def initialize(url, strategy, event_loop, locale = I18n.locale)
     Rails.logger.debug { "#{Time.zone.now.getutc.inspect}: Opening connection to #{url}" }
-    @socket = Faye::WebSocket::Client.new(url, [], strategy.class.websocket_header)
+
+    sentry_transaction = Sentry.get_current_scope&.get_span
+    sentry_span = sentry_transaction&.start_child(op: SENTRY_OP_NAME, start_timestamp: Sentry.utc_now.to_f)
+    http_headers = strategy.class.websocket_header.merge sentry_trace_header(sentry_span)
+
+    @socket = Faye::WebSocket::Client.new(url, [], http_headers)
     @strategy = strategy
     @status = :new
     @event_loop = event_loop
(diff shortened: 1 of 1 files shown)


mpass99 (approved):
Looks great! I really like the nested performance view in Sentry!

lib/runner/connection.rb
```diff
@@ -208,4 +220,51 @@ def handle_timeout(_event)
...
+
+    {
+      headers: http_headers,
+    }
+  end
+
+  def end_sentry_span(sentry_span, event)
+    return unless sentry_span
+
+    sentry_span.set_description("WebSocket #{@socket.url}")
```
mpass99, line 251: As the `@socket.url` contains the runner id and execution id the Sentry Table `Suspect Spans` cannot correctly aggregate the spans (but counts every span individually). We may avoid those dynamic values

mpass99:
Sentry Dashboard

Can we remove either the `http.server` or the `view.process_action.action_controller` span in order to have a correct percentage in the view above?

lib/runner/connection.rb
MrSerth: Thanks for identifying that. I see three possibilities to continue with:

1. Remove the Runner ID (path) and Execution ID (Query)
2. Only remove the Execution ID (Query); that's how Sentry works for the patch route
3. Remove nothing, as the query parameter would only be removed to ensure that no personal identifiable information is sent (which is not the case here).

I compared the WebSocket spans with the `PATCH files` routes, and for the latter (where we don't adjust anything so far), the "Suspect Spans" are also not grouped correctly (only for the runners), not overall for the API route. Did you notice that as well?

MrSerth:
> Can we remove either the `http.server` or the `view.process_action.action_controller` span in order to have a correct percentage in the view above?

Mh, I see what you are referring to. Unfortunately, I don't see a proper way to remove any of these. In my understanding, they represent two different aspects: The `http.server` is, similar to Go, always the incoming request. Then, once handled by our application, those are recorded with the `view.process_action.action_controller`. The fact that all (other) traces are nested there, too, was just changed about three weeks ago (see issue `https://github.com/getsentry/sentry-ruby/issues/1899` and the attached PR). If we had some other processing besides the controller action, this would be listed there as well.

Regarding the numbers at the top, I have the following explanation: The milliseconds are the total of spans having that name including all children (what makes sense, I think). For the percentages, however, I don't know how they are calculated and would assume there's a mistake on their side (we don't do something special there; unusual).

lib/runner/connection.rb
mpass99: Yes, you are right, this problem does also affect the PATCH route.
Although, I see the problem in Sentry and not in our usage I would think about fixing it to have the useful overview of which span uses the most time.
Here, we can fix it by your option 1.
For the other routes by code such as:
``` go
func initSentry(options *sentry.ClientOptions) {
	options.BeforeSendTransaction = func(event *sentry.Event, _ *sentry.EventHint) *sentry.Event {
		fields := strings.Split(strings.Split(strings.Fields(event.Transaction)[1], "?")[0], "/")
		event.Transaction = fields[len(fields)-1]
		return event
	}
	if err := sentry.Init(*options); err != nil {
		log.Errorf("sentry.Init: %s", err)
	}
}
```
What do you think?

PR #1543, opened 2023-02-15
Slightly improve performance for code execution and improve Sentry transaction listing

MrSerth:
This PR addresses first issues we identified with https://github.com/openHPI/poseidon/issues/270 and is a follow-up to https://github.com/openHPI/codeocean/pull/1536#discussion_r1106002703:

* Bulk insert TestrunMessages
* Bulk insert LinterCheckRuns
* Refactor listing files to be independent of exiting runners
* Sentry: Replace UUIDs with * in URLs

Review commit-by-commit is recommended.

PR #1545, opened 2023-02-16
More Sentry monitoring and bug fixes

MrSerth:
Follow-up for #1536 and #1543 

So far, we missed the Sentry instrumentation for score and test runs. See an [old example](https://codeocean.sentry.io/performance/codeocean:e3ff4e7cca03447ba8b502866ae70a15/?environment=production&project=5667283&query=&statsPeriod=7d&transaction=SubmissionsController%23score#span-644dc77b41ac1b2e) and a [new example](https://codeocean.sentry.io/performance/codeocean:4e01343ac3dc482598a49303031f9a42/?project=5667283&query=server_name%3ASebastians-MBP&referrer=performance-transaction-summary&showTransactions=recent&statsPeriod=7d&transaction=SubmissionsController%23score&unselectedSeries=p100%28%29).

If I checked the other spans right, this should be the last one where we miss a lot of instrumentation. I also fixed one bug that occurred recently.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -20,7 +20,7 @@ Metrics/CyclomaticComplexity:
   Max: 20
 
 Metrics/PerceivedComplexity:
-  Max: 23
+  Max: 25
 
 # We don't want to change previous migrations...
 #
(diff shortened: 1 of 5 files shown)


mpass99 (approved):
Good catch with the missing details of the score run!

PR #1698, opened 2023-06-01
LTI replace results on every score

### LINKED ISSUE #1648: LTI: Replace result on every "Score"
Currently, we have two steps for grading a learner's submission: First, a learner clicks "Score" to get feedback and second, a learner clicks "Submit" to send the score to openHPI. This process is cumbersome. We should improve it by sending the result each time a user requested a score.

Initially, scores were only sent once on request (as the LTI parameters were only valid for a single use), but this is not the case any longer. We also reuse the same parameters each time for the RemoteScore action, so the regular browser-based Score should behave similar.

kiragrammel:
Closes #1648


app/assets/javascripts/editor/editor.js.erb
```diff
@@ -769,6 +769,12 @@ var CodeOceanEditor = {
             this.showContainerDepletedMessage();
         } else if (output.status === 'out_of_memory') {
             this.showOutOfMemoryMessage();
+        } else if (output.status === 'scoring_failure') {
```
Dome-GER, line 772: At some point, we might want to switch to `switch...case` 😇

app/controllers/concerns/redirect_behavior.rb
```diff
@@ -55,7 +55,7 @@ def redirect_after_submit
       end
     else
       # redirect to feedback page if score is less than 100 percent
-      if @exercise.needs_more_feedback?(@submission) && !@embed_options[:disable_redirect_to_feedback]
+      if @submission.exercise.needs_more_feedback?(@submission) && !@embed_options[:disable_redirect_to_feedback]
```
Dome-GER, line 58: I assume this concern is used in more places. Do we have the `@submission` available in each and every place?
This looks brittle.

config/locales/de.yml
```diff
@@ -823,7 +825,7 @@ de:
       failure: Fehlerhafte E-Mail oder Passwort.
       success: Sie haben sich erfolgreich angemeldet.
     create_through_lti:
-      session_with_outcome: 'Bitte beachten Sie, dass zur Gutschrift der Punkte Ihr Code nach der Bearbeitung durch Klicken auf den Button "Code zur Bewertung abgeben" eingetragen werden muss.'
+      session_with_outcome: 'Ihre Punkte werden durch Klicken auf den Knopf "Bewerten & Abgeben" übermittelt.'
```
Dome-GER, line 828: "Knopf"? Interesting wording ;)

Dome-GER (commented):
I can only comment on the code, not the domain aspects ;)

app/controllers/concerns/redirect_behavior.rb
kiragrammel: As I understand it, it is ensured that we have a `@submission` by executing `before_action :set_submission, only: %i[redirect_after_submit]` in `submissions_controller.rb`. All other functions in this file are called only in the `redirect_after_submit` function. However, the problem is that rubocop doesn't like `redirect_after_submit` being used in the `submission_controller.rb` because it's not defined there. Since the `RedirectBehavior` is a module, I can't use `before_action` there. Is there any way to solve this? Because `@submission` is needed for the `redirect_after_submit` function. Copying everything from `module RedirectBehavior` in the `submissions_controller.rb` does not seem to be a good solution.

config/locales/de.yml
kiragrammel: Which German word are you using for the English word "button"? "Taste"? Or should I just write: "Ihre Punkte werden durch Klicken auf 'Bewerten & Abgeben' übermittelt."?

app/controllers/concerns/redirect_behavior.rb
MrSerth: Thanks, Dominic, for paying attention. Indeed, we have the concern in more places -- it's also used by the `CommunitySolutionsController`. There, we are also setting the `@submission` variable, so that it can be used safely.

spec/features/editor_spec.rb
```diff
@@ -95,27 +95,18 @@
   end
 
   context 'when an exercise has one or more teacher-defined assessments' do
-    it 'displays the score button' do
+    it 'displays the score and submit button' do
```
MrSerth, line 98: Probably, this change here is unrelated (is it)?

spec/features/editor_spec.rb
```diff
@@ -95,27 +95,18 @@
...
       visit(implement_exercise_path(exercise))
       expect(page).to have_content(exercise.title)
       expect(page).to have_content(I18n.t('exercises.editor.score'))
     end
   end
 
   context 'when an exercise has no teacher-defined assessment' do
-    it 'disables the score button' do
+    it 'disables the score and submit button' do
```
MrSerth, line 106: Probably, this change here is unrelated (is it)?

spec/features/score_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+      let(:sent_score) { 0 }
+    end
+  end
+
+  context 'when LTI outcomes are not supported' do
+    before do
+      allow_any_instance_of(LtiHelper).to receive(:lti_outcome_service?).and_return(false)
+      visit(implement_exercise_path(exercise))
+      click_button(I18n.t('exercises.editor.score'))
+      sleep 10
```
MrSerth, line 109: Have you checked whether `wait_for_ajax` (instead of the `sleep 10`) would also resolve your issue here? This sleep statement does not seem to be very fault-tolerant.

spec/features/score_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+      failed: 0,
+      error_messages: [],
+      passed: 1,
+      score: 1.0,
+      filename: 'index.html_spec.rb',
+      message: 'Well done.',
+      weight: 2.0,
+    }]
+  end
+  let(:scoring_response_zero_points) do
```
MrSerth, line 26: Indeed, the response below does not grant any good score. However, it's also not a real example of a failed test. The following would be, but of course is too complex. Maybe you can construct a nice simplified example, suitable for the given test?

Besides adjusting, you could also consider moving the scoring responses to a factory.

```json
{
    "file_role": "teacher_defined_test",
    "waiting_for_container_time": 0.149172,
    "stdout": "javac -encoding utf8 -cp .:/usr/java/lib/hamcrest-core-1.3.jar:/usr/java/lib/junit-4.13.jar HalloWeltTest2.java\r\njava -cp .:/usr/java/lib/hamcrest-core-1.3.jar:/usr/java/lib/junit-4.13.jar org.junit.runner.JUnitCore HalloWeltTest2\r\nJUnit version 4.13.2\r\n.E.\r\nTime: 0.004\r\nThere was 1 failure:\r\n1) checkForCorrectOutput(HalloWeltTest2)\r\norg.junit.ComparisonFailure: Es wird ein falscher String ausgegeben expected:<willkommen detektiv[ duke]!\r\n> but was:<willkommen detektiv[]!\r\n>\r\n\tat org.junit.Assert.assertEquals(Assert.java:117)\r\n\tat HalloWeltTest2.checkForCorrectOutput(HalloWeltTest2.java:37)\r\n\r\nFAILURES!!!\r\nTests run: 2,  Failures: 1\r\n\r\n",
    "stderr": "make: *** [Makefile:7: test] Error 1\r\n",
    "exit_code": 2,
    "container_execution_time": 0.509629,
    "status": "failed",
    "count": 2,
    "failed": 1,
    "error_messages": [
        "Es wird ein falscher String ausgegeben expected:<willkommen detektiv[ duke]!\r\n> but was:<willkommen detektiv[]!\r\n>\r\n"
    ],
    "passed": 1,
    "score": 0.5,
    "filename": "HalloWeltTest2.java",
    "message": "Die Ausgabe weicht von “Willkommen Detektiv Duke!” ab. Überprüfe den Inhalt von System.out.println() auf Schreibfehler\n",
    "weight": 2.5
}
```

spec/features/score_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+
+        it 'shows a full score notification' do
+          # Text needs to be split because it includes the embedded url in Html which is not shown in the notification
+          # We only compare the shown notification text
+          expect(page).to have_content(I18n.t('exercises.submit.full_score').split('.').first)
+          expect(page).to have_link(nil, href: redirect_after_submit_submission_path(submission))
+        end
+      end
+
+      context 'when scoring is to late' do
```
MrSerth, line 77: Here and in the following:

```suggestion
      context 'when scoring is too late' do
```

spec/policies/exercise_policy_spec.rb
```diff
@@ -186,6 +186,7 @@
     end
   end
 
+  # TODO: Remove?
```
MrSerth, line 189: Yes :)

spec/features/score_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+      visit(implement_exercise_path(exercise))
+      allow(Submission).to receive(:find).and_return(submission)
+      allow(submission).to receive(:calculate_score).and_return(scoring_response)
+      allow(submission).to receive(:normalized_score).and_return(normalized_score)
+      allow_any_instance_of(SubmissionsController).to receive(:send_score).and_return(status:, score_sent: sent_score)
+      click_button(I18n.t('exercises.editor.score'))
+    end
+
+    context 'when the full score is reached' do
+      let(:submission) { create(:submission, exercise:, user:, score: 1.0) }
```
MrSerth, line 61: Since the normalized score is, in this case, always equal to the actual submission score, you could consider renaming the variable below and (maybe more important) use this "score" variable for this submission. Then, the submission might be moved to an outer context, so that only the score needs to be adjusted in the various specs.

spec/features/score_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+        let(:normalized_score) { 0.7 }
+        let(:status) { 'success' }
+
+        it 'shows a scoring to late notification' do
+          expect(page).to have_content(I18n.t('exercises.submit.too_late'))
+        end
+      end
+
+      context 'when scoring fails' do
+        let(:status) { 'error' } # unsupported also works
```
MrSerth, line 87: > unsupported also works

Sure, but shouldn't this be tested in the `context 'when LTI outcomes are not supported'` part below?

In this case, I would also question that `it [should] 'show a scoring failure notification'`. If scoring is not available (for example, if the course on openHPI is in self-paced mode), I probably don't want to receive such a failure notification on every "Score" run, do I? In my local tests, this is also not the case (and thus should be tested).

app/assets/javascripts/editor/editor.js.erb
```diff
@@ -828,6 +840,28 @@ var CodeOceanEditor = {
         });
     },
 
+    showScoringFailureMessage: function () {
+        $.flash.warning({
+            icon: ['fa-solid', 'fa-triangle-exclamation'],
+            text: I18n.t('exercises.submit.failure')
```
MrSerth, line 846: What do you think: Should we extend this message a little to inform the user that the scoring itself was performed successfully but "just" the score has not been transmitted successfully (and that they should retry later again)?

app/assets/javascripts/editor/editor.js.erb
```diff
@@ -828,6 +840,28 @@ var CodeOceanEditor = {
...
+        });
+    },
+
+    showScoringTooLateMessage: function () {
+        $.flash.warning({
+            icon: ['fa-solid', 'fa-triangle-exclamation'],
+            text: I18n.t('exercises.submit.too_late')
```
MrSerth, line 853: We might want to update this text as well, making it a bit more verbose. It's not just about receiving the submission, but also about scoring it and transmitting this score to the e-learning platform. However, we haven't transmitted the achieved score, but rather X points.

app/assets/javascripts/editor/submissions.js
```diff
@@ -198,28 +198,6 @@ CodeOceanEditorSubmissions = {
     }
   },
 
-  submitCode: function(event) {
```
MrSerth, line 201: This method is still used by the community solution (`app/assets/javascripts/community_solution.js:23`), so that the removal will break this part. Rather than just keeping it, we might also consider moving this code closer to the community solution (since it is used there).

MrSerth (commented):
Congratulations on this PR, it was a really though one and touched many different parts of the system. Here is a first review with my feedback. It looks like many comments, but I actually don't think there are so many -- just a few renamings and textual adjustments here and there. 

While taking about textual adjustments: Would you please be kind enough to have another look at the translations (in `en.yml` and `de.yml`)? I have the feeling that some translations listed there are no longer needed after your PR:

  - The former tooltip shown for the submit button can be removed
  - The texts for the different submit buttons should be removed (and the according hints should be updated, this is a comment I also made somewhere else during this review)
  - There is no longer a confirmation for submit actions (an old translation not used any more since quite a while).
  - There is a comment in lti.rb in / around method `clear_lti_session_data`, that could be updated.

config/locales/de.yml
```diff
@@ -531,6 +531,7 @@ de:
     submit:
       failure: Beim Übermitteln Ihrer Punktzahl ist ein Fehler aufgetreten. Bitte versuchen Sie es später erneut.
       too_late: Ihre Abgabe wurde erfolgreich gespeichert, ging jedoch nach der Abgabefrist ein.
+      full_score: Herzlichen Glückwunsch! Sie haben die maximale Punktzahl für diese Aufgabe erreicht. Klicken Sie <a href="%{url}" rel="nofollow">hier</a>, um die Aufgabe abzuschließen.
```
MrSerth, line 534: Please add the `class="alert-link"` to the link, as this will be required with the newest style changes.

app/assets/javascripts/exercise_graphs.js
```diff
@@ -123,6 +123,9 @@ $(document).on('turbolinks:load', function() {
             if(largestSubmittedTimeStamp.cause === "assess"){
                 largestArrayForRange = submissionsScoreAndTimeAssess;
                 x.domain([0,largestArrayForRange[largestArrayForRange.length - 1][1]]).clamp(true);
+            } else if(largestSubmittedTimeStamp.cause == "submit"){
+                largestArrayForRange = submissionsScoreAndTimeSubmits;
+                x.domain([0,largestArrayForRange[largestArrayForRange.length - 1][1]]).clamp(true);
```
MrSerth, line 128: These lines seem to be duplicated, but shouldn't be inserted here.

spec/controllers/exercises_controller_spec.rb
```diff
@@ -288,95 +288,6 @@
     end
   end
 
-  describe 'POST #submit' do
```
MrSerth, line 291: Rather than removing this test, we could consider transforming it for our new `SubmissionsController#finalize` method.

app/assets/javascripts/community_solution.js
```diff
@@ -21,8 +21,29 @@ $(document).on('turbolinks:load', function() {
...
+        $('#accept').one('click', submitCode.bind(CodeOceanEditor));
     }
-
 });
+
+function submitCode(event) {
+    const button = $(event.target) || $('#submit');
+    this.startSentryTransaction(button);
+    this.teardownEventHandlers();
+    this.createSubmission(button, null, function (response) {
```
MrSerth, line 33: Have you tested that these method calls are still successful? I would doubt that, since the `this.` reference changed.

MrSerth:
Since we changed many files in the default branch, we should continue here with a thought-through rebase and necessary adjustments.

config/locales/en.yml
```diff
@@ -530,8 +525,9 @@ en:
       external_users: External Users
       finishing_rate: Finishing Rate
     submit:
-      failure: An error occurred while transmitting your score. Please try again later.
-      too_late: Your submission was saved successfully but was received after the deadline passed.
+      failure: The scoring was performed successfully but an error occurred while transmitting your score. Please try again later.
+      too_late: Your submission was saved successfully but was received after the deadline passed so that only %{score_sent} points were transmitted.
+      full_score: Congratulations! You have achieved the highest possible score for this exercise. Click <a href="%{url}" class="alert-link">here</a> to complete the exercise.
```
MrSerth, line 530: I would suggest highlighting the link a bit more, so that it's clearer to find. 

```suggestion
      full_score: Congratulations! You have achieved the highest possible score for this exercise. <a href="%{url}" class="alert-link">Please click here to finish the exercise now.</a>
```

config/locales/en.yml
```diff
@@ -530,8 +525,9 @@ en:
       external_users: External Users
       finishing_rate: Finishing Rate
     submit:
-      failure: An error occurred while transmitting your score. Please try again later.
-      too_late: Your submission was saved successfully but was received after the deadline passed.
+      failure: The scoring was performed successfully but an error occurred while transmitting your score. Please try again later.
+      too_late: Your submission was saved successfully but was received after the deadline passed so that only %{score_sent} points were transmitted.
```
MrSerth, line 529: Potentially, we can simplify that a little:

```suggestion
      too_late: Your submission was saved successfully but was received after the deadline, so that only %{score_sent} points were transmitted.
```

config/locales/en.yml
```diff
@@ -822,7 +818,7 @@ en:
       failure: Invalid email or password.
       success: Successfully signed in.
     create_through_lti:
-      session_with_outcome: 'Please click "Submit Code for Assessment" after scoring to send your score to %{consumer}.'
+      session_with_outcome: 'By clicking on "Score", your points will automatically be added to your progress.'
```
MrSerth, line 821:
```suggestion
      session_with_outcome: 'By clicking on "Score", your points will be added automatically to your progress.'
```

config/locales/de.yml
```diff
@@ -530,8 +525,9 @@ de:
       external_users: Externe Nutzer
       finishing_rate: Abschlussrate
     submit:
-      failure: Beim Übermitteln Ihrer Punktzahl ist ein Fehler aufgetreten. Bitte versuchen Sie es später erneut.
-      too_late: Ihre Abgabe wurde erfolgreich gespeichert, ging jedoch nach der Abgabefrist ein.
+      failure: Die Bewertung wurde erfolgreich durchgeführt, aber beim Übermitteln Ihrer Punktzahl ist ein Fehler aufgetreten. Bitte versuchen Sie es später erneut.
+      too_late: Ihre Abgabe wurde erfolgreich gespeichert, ging jedoch nach der Abgabefrist ein, sodass nur %{score_sent} Punkte übertragen wurden.
+      full_score: Herzlichen Glückwunsch! Sie haben die maximale Punktzahl für diese Aufgabe erreicht. Klicken Sie <a href="%{url}" class="alert-link">hier</a>, um die Aufgabe abzuschließen.
```
MrSerth, line 530:
```suggestion
      full_score: Herzlichen Glückwunsch! Sie haben die maximale Punktzahl für diese Aufgabe erreicht. <a href="%{url}" class="alert-link">Klicken Sie hier, um die Aufgabe jetzt abzuschließen.</a>
```

app/views/exercises/_editor_output.html.slim
```diff
@@ -59,15 +59,10 @@ div.d-grid id='output_sidebar_uncollapsed' class='d-none col-sm-12 enforce-botto
         .progress
           .progress-bar role='progressbar'
 
-      br
-      - if lti_outcome_service?(@exercise.id, external_user_id)
-        p.text-center = render('editor_button', classes: 'btn-lg btn-success d-none', data: {:'data-url' => submit_exercise_path(@exercise)}, icon: 'fa-solid fa-paper-plane', id: 'submit', label: t('exercises.editor.submit'))
         - if @exercise.submission_deadline.present? || @exercise.late_submission_deadline.present?
           #deadline data-submission-deadline=@exercise.submission_deadline&.rfc2822 data-late-submission-deadline=@exercise.late_submission_deadline&.rfc2822
             h4 = t('exercises.editor.deadline')
             = t('exercises.editor.hints.disclaimer')
```
MrSerth, line 65: I might be wrong, but I would have expected that the indention level of these lines should change (move more to the left). Have you verified everything is working here correctly?

app/controllers/submissions_controller.rb
```diff
@@ -72,6 +73,11 @@ def download_file
     end
   end
 
+  def finalize
+    @submission.update(cause: 'submit')
```
MrSerth, line 77: Since we have no other guard clause here, I would use the bang method:

```suggestion
    @submission.update!(cause: 'submit')
```

app/controllers/submissions_controller.rb
```diff
@@ -252,14 +258,33 @@ def score
...
+          client_socket&.send_data(JSON.dump({cmd: :status, status: :scoring_too_late, score_sent: response[:score_sent]}))
+        end
+      else
+        client_socket&.send_data(JSON.dump({cmd: :status, status: :scoring_failure}))
+      end
+    end
+
+    if @submission.score == @submission.exercise.maximum_score
+      @url = finalize_submission_path(@submission)
+      client_socket&.send_data(JSON.dump({cmd: :status, status: :full_score_reached, url: @url}))
```
MrSerth, line 278: I know that we have heavily reworked this part previously. Since the new architecture is already more flexible, we can probably rethink these lines and just "add" the part to send a JSON through the WebSocket connection there.

_This shouldn't be a dedicated comment, it's more a reminder about a nice architecture._

app/controllers/submissions_controller.rb
```diff
@@ -252,14 +258,33 @@ def score
...
+    end
+
+    kill_client_socket(client_socket)
```
MrSerth, line 281: Is there a special reason you moved the `kill_client_socket(client_socket)` from the `ensure` block? Or is it just a render error based on the conflicts?

spec/policies/exercise_policy_spec.rb
```diff
@@ -186,26 +186,6 @@
     end
   end
 
-  permissions :submit? do
```
MrSerth, line 189: Deleting this policy is okay, but we should adjust the specs for the `SubmissionPolicy` to test for the newly-introduced `finalize?` as well.

MrSerth (commented):
Nice done! I've just rechecked the code and added a few minor comments. Probably, it takes the most time to integrate these changes with the existing ones, resolving the merge conflicts and cleaning the history. Otherwise, I think it looks mostly good 🏕️

app/views/exercises/_editor_output.html.slim
kiragrammel: It looks the same when moving it two spaces left. But I don't know when the expression is true. When I set a submission deadline for the exercise on openHPI it is not shown.

MrSerth: You need to specify a submission deadline within the exercise on CodeOcean -- not on openHPI. This is completely independent of any LTI information.

When reading the code (but without trying live), I have the feeling that the deadline information is nested under the `<div id='score'>`, which should not be the case. Moving those lines to the left would have them on the same level as the `<div id='score'>`, not nested below.

config/locales/en.yml
```diff
@@ -560,8 +555,10 @@ en:
       programming_groups: Programming Groups
       finishing_rate: Finishing Rate
     submit:
-      failure: An error occurred while transmitting your score. Please try again later.
-      too_late: Your submission was saved successfully but was received after the deadline passed.
+      failure: The scoring was performed successfully but an error occurred while transmitting your score. Please try again later.
+      too_late: Your submission was saved successfully but was received after the deadline, so that only %{score_sent} points were transmitted.
```
kiragrammel, line 559: Should we add the redirect option `<a href="%{url}" class="alert-link">Please click here to finish the exercise now.</a>` here and also for `too_late_and_not_for_all_users_submitted`?

app/controllers/submissions_controller.rb
```diff
@@ -275,7 +275,17 @@ def score
 
     # Submit the score
     if @submission.users.map {|user| lti_outcome_service?(@submission.exercise, user, @submission.study_group_id) }.any?
-      transmit_lti_score
+      url = finalize_submission_path(@submission)
+      submit_info = transmit_lti_score
+      submit_info&.[](:status)&.each do |s|
```
MrSerth, line 280: Why are you iterating over the `status` here? If I read the code correctly, the status never contains more than one element, does it?

For reference: In `transmit_lti_score`, you're creating a new, empty array for the status and are later on pushing symbols to it. However, the pushing of a status is performed outside a loop and conditional as part of an `if-else` statement, avoiding two status information to be pushed. Further, I was not able to reproduce any code path for this case.

app/controllers/submissions_controller.rb
```diff
@@ -275,7 +275,17 @@ def score
...
+        client_socket&.send_data(JSON.dump({
+          cmd: :status,
+          status: s,
```
MrSerth, line 283: Do you think it would be easier to send up to two distinct messages instead of one for LTI success / failure and on time / after the deadline? 

The LTI messages could be translated to the following LTI styles:
- success (all successful)
- warning (some failed, some worked)
- danger (all failed)

For the deadline, we could:
- not show any message if the deadline has been fulfilled
- an info if the deadline has passed (i.e., the score is reduced somehow). Here, I originally thought of having another warning (yellow), but this doesn't play nice with the LTI message styles (we can only display one warning at a time). Therefore, I think the info (blue) might be the second-best option currently.

Before changing anything here, do you think this change would make sense? Is it reducing the overall complexity / making the code easier to maintain (because those two information are no longer "merged") or do you have other ideas?

config/locales/en.yml
MrSerth: Ideally, I would only add this link iff the user seeing the flash message was the one for which scoring worked. Currently, this case is not considered, as shown with the following screenshot (of course, that's me!):

Bildschirmfoto 2023-10-30 um 12 38 01

app/controllers/submissions_controller.rb
kiragrammel: Yes, probably you are right that this change would reduce some complexity. Then we only have three messages + sometimes an additional one and it also is easier to maintain the messages. 
I am just not sure what we should write in the messages if the full score is reached but after the deadline because then the points are reduced. Maybe we then write in the success message only that the exercise was solved completely without saying anything about the points and then in the additional message for the too late submission that the points are reduced?

MrSerth: Yeah, I like that! 👍  This means, we are further decoupling _finishing_ the exercise (in terms of solving it to the satisfaction of the unit tests) and _scoring_ it (which could be influenced by the deadline).

config/locales/en.yml
kiragrammel: Can one of these errors `IMS::LTI::XMLParseError, Net::OpenTimeout, Net::ReadTimeout, Errno::ECONNRESET, SocketError, EOFError` also occur if the lti_outcome_service is available?

MrSerth: Those errors can _only_ occur if the `lti_outcome_service` is available. They can happen if the LMS is unavailable or returned an invalid XML `IMS::LTI::XMLParseError`...

app/controllers/concerns/redirect_behavior.rb
```diff
@@ -24,7 +24,7 @@ def redirect_after_submit
         rfc = @submission.own_unsolved_rfc(current_user)
         if rfc
           # set a message that informs the user that his own RFC should be closed.
-          flash[:notice] = I18n.t('exercises.submit.full_score_redirect_to_own_rfc')
+          flash[:notice] = I18n.t('exercises.submit.exercise_finished_redirect_to_own_rfc')
```
MrSerth, line 27: Nice, thanks for updating the translation here as well 👍

app/controllers/submissions_controller.rb
```diff
@@ -279,20 +279,18 @@ def score
     # To enable hints when scoring a submission, uncomment the next line:
     # send_hints(client_socket, StructuredError.where(submission: @submission))
 
-    # Submit the score
-    if @submission.users.map {|user| lti_outcome_service?(@submission.exercise, user, @submission.study_group_id) }.any?
-      url = finalize_submission_path(@submission)
-      submit_info = transmit_lti_score
-      submit_info&.fetch(:status, [])&.each do |s|
+    # The own score can only be transmitted when the lti_outcome_service is available
+    # if it is not available the user also does not send the score for the programming group members
```
MrSerth, line 283: Suggestion (line could be too long and might require another break):

```rb
# A score can only be transmitted if the `lti_outcome_service` has been announced during the LTI exercise launch.
# Since the absence of the `lit_outcome_service` for the **current user** is considered a permanent error (i.e., because the deadline on Xikolo as an LMS passed), we also avoid sending a score for any (other) member of the programming group.
```

app/controllers/submissions_controller.rb
```diff
@@ -509,39 +507,42 @@ def set_testrun
...
-
-    responses.each do |response|
-      if Lti::ERROR_STATUS.include? response[:status]
-        failed_users << response[:user]
-      elsif response[:score_sent] != @submission.normalized_score # the score was sent successfully, but received too late
-        too_late = true
-        score_sent = response[:score_sent]
-      end
+  def check_scoring_too_late(submit_info, client_socket)
+    if submit_info[:deadline] == :within_grace_period || submit_info[:deadline] == :after_late_deadline
```
MrSerth, line 511: This could be written as:

```rb
if %i[within_grace_period after_late_deadline].include? submit_info[:deadline]
```

app/controllers/submissions_controller.rb
```diff
@@ -509,39 +507,42 @@ def set_testrun
...
-    elsif too_late # submissions are too late
-      status.push(:scoring_too_late)
-      {status:, score_sent:}
-    elsif @submission.score == @submission.exercise.maximum_score
-      status.push(:full_score_reached)
-      {status:}
+  def transmit_lti_score(client_socket)
+    submit_info = send_scores(@submission)
+
+    if (submit_info[:users][:all].count == submit_info[:users][:success].count) & (submit_info[:score][:original] == BigDecimal('1.0'))
```
MrSerth, line 523: I am not quite sure whether the operator `&` is correct here: Are you intending to use a logical AND?

MrSerth, line 523: Further, the `BigDecimal`  could be written as `1.to_d` (or `1.0.to_d`).

MrSerth, line 523: Here, it might make sense to use a case-when construct. This could allow to clean up the code and will probably also avoid redoing the same calculations over and over again (in the current implementation, the `submit_info[:users][:all].count` is performed multiple times, even though the value is unlikely to change between the cases.

app/controllers/submissions_controller.rb
```diff
@@ -279,20 +279,18 @@ def score
...
-    if @submission.users.map {|user| lti_outcome_service?(@submission.exercise, user, @submission.study_group_id) }.any?
-      url = finalize_submission_path(@submission)
-      submit_info = transmit_lti_score
-      submit_info&.fetch(:status, [])&.each do |s|
+    # The own score can only be transmitted when the lti_outcome_service is available
+    # if it is not available the user also does not send the score for the programming group members
+    if lti_outcome_service?(@submission.exercise, @current_user, @submission.study_group_id)
+      transmit_lti_score(client_socket)
+    else
+      if @submission.score == @submission.exercise.maximum_score
```
MrSerth, line 287: This condition is semantically similar to the condition below in the `transmit_lti_score`. Is there a mechanism to avoid the duplication (of the lines below with the `client_socket&.send_data` for `:exercise_finished` in `transmit_lti_score`?

app/controllers/submissions_controller.rb
```diff
@@ -509,39 +507,42 @@ def set_testrun
...
-    score_sent = nil
-
-    responses.each do |response|
-      if Lti::ERROR_STATUS.include? response[:status]
-        failed_users << response[:user]
-      elsif response[:score_sent] != @submission.normalized_score # the score was sent successfully, but received too late
-        too_late = true
-        score_sent = response[:score_sent]
-      end
+  def check_scoring_too_late(submit_info, client_socket)
```
MrSerth, line 510: I would be open to have this message appear in all cases of the `transmit_lti_score` method, further reducing the complexity of the code (it is, indeed, independent of the LTI stuff, isn't it?).

MrSerth, line 510: I am not quite sure how I like the method name: "Checking" something would somehow imply we are returning the result of the check to the callee, wouldn't it? Otherwise, it is pretty close to what we do, but rather inform the learner directly. Maybe I have another idea later today...

app/controllers/submissions_controller.rb
```diff
@@ -509,39 +507,42 @@ def set_testrun
...
-      status.push(:scoring_too_late)
-      {status:, score_sent:}
-    elsif @submission.score == @submission.exercise.maximum_score
-      status.push(:full_score_reached)
-      {status:}
+  def transmit_lti_score(client_socket)
+    submit_info = send_scores(@submission)
+
+    if (submit_info[:users][:all].count == submit_info[:users][:success].count) & (submit_info[:score][:original] == BigDecimal('1.0'))
+      client_socket&.send_data({
```
MrSerth, line 524: Probably, I would somehow extract the `client_socket&.send_data` to be outside of the if-else construct. This could be beneficial, since it might fail (we are dealing with some network-related operations), where the other operations are, under regular operation, not expected to fail.

At most, I would return the Hash (not the JSON). Further, I am thinking whether there is more potential to reduce the code, i.e., by [adding the status later through a `merge`](https://ruby-doc.org/3.2.2/Hash.html#method-i-merge) (avoiding the repetition here in each case).

MrSerth (commented):
Nice done, I like the integration of those changes and the refactoring you performed very much. Still, as you know me already, I have further ideas, which I just added as comments 💬.

config/locales/en.yml
kiragrammel: Question has been clarified, since now the reaching of the full score is independent of transmission errors. Also, we check if it is the current_user where errors occur.

app/controllers/submissions_controller.rb
kiragrammel: Now we check it differently in the submissions model.

kiragrammel: The conditions changed so that it is now easier to use an if-else construct.

app/controllers/submissions_controller.rb
kiragrammel: We talked about it that the notification is **not** shown when a failure occurred because then it is uninteresting if the score was reduced.

app/controllers/submissions_controller.rb
MrSerth: Probably not. However, we can also remove the "old" comment (will do so).

MrSerth (commented):
I reviewed the changes once again and made a few adjustments where appropriate. Other than that, I think that the actual code is now fine (I've tested that in multiple constellations) and am also pretty happy with the code that we've come up with. From my point of view, the code quality has dramatically improved through the various iteration cycles we had and is also a much needed improvement over the original structure.

The remaining part of this PR is mainly to have a look at the specs. First, the existing (failing) specs need to be fixed, and second, we should carefully add more specs where needed.

spec/controllers/exercises_controller_spec.rb
kiragrammel: Reused in `score_spec.rb`

spec/features/score_spec.rb
```diff
@@ -0,0 +1,286 @@
...
+      users: {all: users_success + users_error + users_unsupported, success: users_success, error: users_error, unsupported: users_unsupported},
+      score: {original: score, sent: score_sent},
+      deadline:,
+      detailed_results: [],
+    }
+  end
+
+  let(:calculate_response) do
+    {
+      status:,
```
MrSerth, line 20: In the case `when LTI outcomes are not supported`, you're setting this `status` to `unsupported`. However, I am not sure where this could ever happen within the actual code (when not using a mock). Do you have a pointer for me, please?

kiragrammel: It seems to be a mistake from my side. There only is `ok` and `failed` as status, right? In this unsupported case is the status then failed?

MrSerth: > It seems to be a mistake from my side.

Alright.

> There only is ok and failed as status, right? 

In addition to these two (which I also see), there are a few more, but all related to runner management -- nothing in regard with LTI.

> In this unsupported case is the status then failed?

No. The score run can still succeed (Poseidon, ...), while the LTI submission could be unsupported (missing LTI params). Those are two different things...

I am currently making a few suggestions to your code, including a removal of this `status` variable. Please be patient for my upcoming changes :)

MrSerth (commented):
Awesome changes, thank you!

I've just taken the time to review the `score_spec.rb` and apply some changes there for a smoother definition of the specs with less code repetition. Further, I added some test cases for exception handling. My changes are separated in various commits, to ease reviewing. Please have a look there.

Next to these nice tests, I would suggest adding some additional test cases for the RemoteEvaluationsController and for finalising a submission. 
- For the RemoteEvaluationsController, I would assume that significantly less tests are expected: Using the [`try_lti` method](https://github.com/openHPI/codeocean/blob/581717a652dde6e99240d304286f5d78a9f04140/app/controllers/remote_evaluation_controller.rb#L41) as a base, I would expect those 6 tests + 3 cases in `create_and_score_submission`. I don't expect any (relevant) change to the RemoteEvaluationsController.
- For the `SubmissionsController#finalise`, the situation might be a little more complex, but is mainly testing the [`redirect_after_submit`](https://github.com/openHPI/codeocean/blob/584fe390d82a58961b4dcc5d17e55dd4d5b68ae4/app/controllers/concerns/redirect_behavior.rb#L6). Within the `redirect_after_submit`, I think we still have "dead code" that cannot be reached any longer (it should be removed, of course).

spec/concerns/redirect_behavior_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    context 'when community solution is submitted' do
+      before do
+        allow(described_class).to receive(redirect_to_community_solution?).and_return(true)
+      end
+
+      it 'redirects to community solution' do
+        expect_redirect { edit_community_solution_path(community_solution, lock_id: community_solution_lock.id) }
+        # expect(controller).to receive(:redirect_to).with(edit_community_solution_path(community_solution, lock_id: community_solution_lock.id))
+      end
+    end
```
MrSerth, line 30: You will probably need to move all tests to `specs/submissions_controller_spec.rb`. Then, this test would read as:

```suggestion
    context 'when a community solution is offered' do
      let!(:community_solution) { CommunitySolution.create(exercise:) }

      before do
        allow(Java21Study).to receive(:allow_redirect_to_community_solution?).and_return(true)
        post :finalize, params: {id: submission.id}
      end

      expect_redirect { edit_community_solution_path(community_solution, lock_id: CommunitySolutionLock.last) }
    end
```

MrSerth (commented):
I've finished my review for the RemoteEvaluationsController specs, and also paved the way for the upcoming RedirectBehavior specs a little.

MrSerth (approved):
Awesome, nicely done and many thanks for your perseverance! 👏 🥳 🎉 🥂 👯 

I am very that you took the challenge to work on this crucial part of the application while digging deep into everything that's LTI related and covering your changes with so many, extensive tests. As you see, almost each line touched is covered, and a magnitude more lines are newly covered by tests even without any change.

Furthermore, I just finished my review of all code parts and am really satisfied with the result. For a final test, I've deployed the changes to Staging and CodeMoon, allowing us to test-drive the changes in a real-world context. 

When doing so myself, I discovered another error with the user exercise feedbacks, not caused by our recent change. Besides that, I was surprised about the 10% rule of the feedbacks, because it was much more frequent in my tests. Regardless of that, the basic workflow was as expected. Nevertheless, it probably makes sense to have a final check 🤓.

spec/controllers/submissions_controller_spec.rb
```diff
@@ -6,7 +6,8 @@
   render_views
```
kiragrammel, line 6: This line is probably not needed as it is written a second time in line 280 for the 'GET #show.json' tests. For me locally the tests pass without it.

spec/controllers/submissions_controller_spec.rb
```diff
@@ -98,6 +108,85 @@
...
+      context 'with an RfC' do
+        before do
+          rfc.save!
+          allow_any_instance_of(Submission).to receive(:redirect_to_feedback?).and_return(false)
+          perform_request.call
+        end
+
+        context 'when an own RfC is unsolved' do
+          let(:rfc) { create(:rfc, user: current_user, exercise:, submission:) }
+          let(:flash_message) { I18n.t('exercises.editor.exercise_finished_redirect_to_own_rfc') }
```
kiragrammel, line 166: This variable is not used so the line probably could be removed. For me locally the tests pass without the line.

PR #1843, opened 2023-08-20
Synchronize text in editor

kiragrammel:
What does this change?
* Use Action Cable to send changes in editor to programming group team members
* Changes are displayed in the editor so that all members of a programming group see the same written code

Decisions/Choices I made:
* Client side decision if changes are own changes or from team members instead of server side decision
   - In the case of a server-side decision the changes are send to all members of a programming group except one self. For that a new channel for each team member with the server is needed. Therefore, the decision was made that the work is done on client side to reduce the number of channels.

### DIFF
--- a/app/assets/javascripts/channels/synchronized_editor_channel.js
+++ b/app/assets/javascripts/channels/synchronized_editor_channel.js
@@ -1,0 +1,38 @@
+$(document).on('turbolinks:load', function () {
+
+  if (window.location.pathname.includes('/implement')) {
+    const editor = $('#editor');
+    const exercise_id = editor.data('exercise-id');
+    const current_user_id = editor.data('user-id');
+    const current_contributor_id = editor.data('contributor-id');
+
+    if ($.isController('exercises') && current_user_id !== current_contributor_id) {
+
+      App.synchronized_editor = App.cable.subscriptions.create({
+        channel: "SynchronizedEditorChannel", exercise_id: exercise_id
+      }, {
+
+
+        connected() {
+          // Called when the subscription is ready for use on the server
+        },
+
+        disconnected() {
+          // Called when the subscription has been terminated by the server
+        },
+
+        received(data) {
+          // Called when there's incoming data on the websocket for this channel
+          if (current_user_id !== data['current_user_id']) {
+            CodeOceanEditor.applyChanges(data['delta']['data']);
+          }
+        },
+
+        send_changes(delta) {
+          const delta_with_user_id = {current_user_id: current_user_id, delta: delta}
+          this.perform('send_changes', {delta_with_user_id: delta_with_user_id});
+        }
+      });
+    }
+  }
+});
--- a/app/assets/javascripts/editor/editor.js.erb
+++ b/app/assets/javascripts/editor/editor.js.erb
@@ -29,6 +29,7 @@ var CodeOceanEditor = {
     running: false,
 
     lastCopyText: null,
+    lastDeltaObject: null,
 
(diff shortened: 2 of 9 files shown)


app/assets/javascripts/channels/synchronized_editor_channel.js
```diff
@@ -18,11 +20,14 @@ $(document).on('turbolinks:load', function() {
...
+        if (current_user_id !== data['current_user_id']) {
+          CodeOceanEditor.applyChanges(data['delta']['data']);
+        }
       },
 
-      speak(data) {
-        this.perform('speak', { message: data });
+      send_changes(delta) {
+        const delta_with_user_id = {current_user_id: current_user_id, delta: delta}
+        this.perform('send_changes', { delta_with_user_id: delta_with_user_id});
```
MrSerth, line 30: We don't have a linter in JavaScript yet. Therefore, we should try to fix the code ourselves (or at least use IntelliJ's code formatting with <kbd>command</kbd> + <kbd>option</kbd> + <kbd>L</kbd>).

app/assets/javascripts/editor/editor.js.erb
```diff
@@ -1019,6 +1022,25 @@ var CodeOceanEditor = {
...
+            return false;
+        }
+
+        const delta_data = delta.data
+        return delta_data.action === last_delta.action &&
+          delta_data.range.start.row === last_delta.range.start.row &&
+          delta_data.range.start.column === last_delta.range.start.column &&
+          delta_data.range.end.row === last_delta.range.end.row &&
+          delta_data.range.end.column === last_delta.range.end.column &&
+          delta_data.text === last_delta.text;
```
MrSerth, line 1041: I am aware that this is not ideal yet, but am fine to have it that way. Maybe just add a small comment above why we are making this very manual comparison :+1:.

MrSerth (commented):
Awesome, thanks! That's really a nice work, and I am more than confident that we will be able to merge that very soon! 

Please find below a few comments which I noticed while going through your changes. After some rebasing / cherry-picking and potentially squashing, only a few comments should be left, so that we can then proceed further.

app/channels/synchronized_editor_channel.rb
```diff
@@ -0,0 +1,21 @@
...
+    stop_all_streams
+  end
+
+  def specific_channel
+    reject unless ProgrammingGroupPolicy.new(current_user, programming_group).stream_sync_editor?
+    "synchronized_editor_channel_group_#{programming_group.id}"
+  end
+
+  def programming_group
+    @programming_group ||= current_user.programming_groups.find_by(exercise_id: params[:exercise_id])
```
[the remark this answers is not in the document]
kiragrammel: I cannot access the session at this position. Maybe because the `SynchronziedEditorChannel` only inherits from `ApplicationCable::Channel`. Adding `include ActionController` and `include ApplicationHelper` also didn't help and is maybe a bit overkill when we only want to get the session data.

MrSerth (approved):
I've made a few more changes, added some tests, rebased, and squashed the changes.

The main change, besides the few policy specs, were 1) some adjustments to the `find_verified_contributor` method. Rather than evaluating the `current_contributor` (which is just the attribute of the class and hence always `nil` first), we rather check for the value directly. 2) The `author_in_programming_group?` checked with `respond_to? :programming_group?` whether a `@record` is a programming group. This is not sufficient, then a user would also respond to that method (we are not considering the return value).

Further, I slightly change the `participate?` method, so that we can enable it on our system more easily.

PR #1850, opened 2023-08-22
Update from proforma to proformaxml

MrSerth:
This PR switches to the official proformaxml gem, as published on RubyGems.org

I also decided to rename the custom error class, besides not required technically. However, I haven't renamed the services, which we could still do.


kkoehn (approved):
Looks good

PR #1866, opened 2023-08-31
Further changes for pair programming

kiragrammel:
* Assign users to A/B groups
* Show in implement route in statusbar if it is a pair programming session and connection status of team partner
* Add button to start (video) chat with programming group
* Add etherpad to exchange user IDs
* Fixed synchronized editor for multiple files (solves #1868)
* Unsubscribe from channel when redirecting
* Limit programming groups to two people
* Etherpad link for pair programming feedback
* Add events for pair programming study

### DIFF
--- a/app/assets/javascripts/channels/synchronized_editor_channel.js
+++ b/app/assets/javascripts/channels/synchronized_editor_channel.js
@@ -24,12 +24,12 @@ $(document).on('turbolinks:load', function () {
         received(data) {
           // Called when there's incoming data on the websocket for this channel
           if (current_user_id !== data['current_user_id']) {
-            CodeOceanEditor.applyChanges(data['delta']['data']);
+            CodeOceanEditor.applyChanges(data['delta']['data'], data['active_file']);
           }
         },
 
-        send_changes(delta) {
-          const delta_with_user_id = {current_user_id: current_user_id, delta: delta}
+        send_changes(delta, active_file) {
+          const delta_with_user_id = {current_user_id: current_user_id, active_file: active_file, delta: delta}
           this.perform('send_changes', {delta_with_user_id: delta_with_user_id});
         }
       });
--- a/app/assets/javascripts/channels/synchronized_editor_channel.js
+++ b/app/assets/javascripts/channels/synchronized_editor_channel.js
@@ -23,13 +23,24 @@ $(document).on('turbolinks:load', function () {
 
         received(data) {
           // Called when there's incoming data on the websocket for this channel
-          if (current_user_id !== data['current_user_id']) {
-            CodeOceanEditor.applyChanges(data['delta']['data'], data['active_file']);
+          if (current_user_id !== data.current_user_id) {
+            switch(data.command) {
+              case 'editor_change':
+                CodeOceanEditor.applyChanges(data.delta.data, data.active_file);
+                break;
+              case 'connection_change':
+                CodeOceanEditor.showPartnersConnectionStatus(data.status, data.current_user_name);
+                this.perform('send_hello');
+                break;
+              case 'hello':
+                CodeOceanEditor.showPartnersConnectionStatus(data.status, data.current_user_name);
+                break;
+            }
           }
         },
 
         send_changes(delta, active_file) {
-          const delta_with_user_id = {current_user_id: current_user_id, active_file: active_file, delta: delta}
+          const delta_with_user_id = {command: 'editor_change', current_user_id: current_user_id, active_file: active_file, delta: delta}
           this.perform('send_changes', {delta_with_user_id: delta_with_user_id});
         }
       });
--- a/app/assets/javascripts/editor/editor.js.erb
+++ b/app/assets/javascripts/editor/editor.js.erb
@@ -1064,6 +1064,9 @@ var CodeOceanEditor = {
         this.initializeDeadlines();
         CodeOceanEditorTips.initializeEventHandlers();
 
+        window.addEventListener("turbolinks:before-render", App.synchronized_editor?.unsubscribe.bind(App.synchronized_editor));
+        window.addEventListener("beforeunload", App.synchronized_editor?.unsubscribe.bind(App.synchronized_editor));
+
         window.addEventListener("turbolinks:before-render", this.autosaveIfChanged.bind(this));
         window.addEventListener("beforeunload", this.autosaveIfChanged.bind(this));
         // create autosave when the editor is opened the first time
--- a/app/assets/javascripts/editor/editor.js.erb
+++ b/app/assets/javascripts/editor/editor.js.erb
@@ -339,7 +339,7 @@ var CodeOceanEditor = {
                     CodeOceanEditor.lastDeltaObject = null;
                     return;
                 }
-                App.synchronized_editor?.send_changes(deltaObject);
+                App.synchronized_editor?.send_changes(deltaObject, this.active_file);
 
                 // TODO: This is a workaround for a bug in Ace. Remove when upgrading Ace.
                 this.handleUTF16Surrogates(deltaObject, session);
@@ -1022,9 +1022,10 @@ var CodeOceanEditor = {
         }
     },
 
-    applyChanges: function (delta) {
+    applyChanges: function (delta, active_file) {
         this.lastDeltaObject = delta;
-        this.editors[0].session.doc.applyDeltas([delta]);
+        const editor = this.editor_for_file.get(active_file.filename)
+        editor.session.doc.applyDeltas([delta]);
     },
 
     compareDeltaObjects: function (delta, last_delta) {
--- a/app/assets/javascripts/editor/editor.js.erb
+++ b/app/assets/javascripts/editor/editor.js.erb
@@ -1044,6 +1044,17 @@ var CodeOceanEditor = {
           delta_data.text === last_delta.text;
     },
 
+    showPartnersConnectionStatus: function (status, username) {
+        switch(status) {
+            case 'connected':
+                $('#pg_session').text(I18n.t('exercises.editor.is_online', {name: username}));
+                break;
+            case 'disconnected':
+                $('#pg_session').text(I18n.t('exercises.editor.is_offline', {name: username}));
+                break;
+        }
+    },
+
     initializeEverything: function () {
         CodeOceanEditor.editors = [];
         this.initializeRegexes();
--- a/app/assets/javascripts/editor/editor.js.erb
+++ b/app/assets/javascripts/editor/editor.js.erb
@@ -382,6 +382,20 @@ var CodeOceanEditor = {
         $(document).on('click', '#results a', this.showOutput.bind(this));
         $(document).on('keydown', this.handleKeyPress.bind(this));
         $(document).on('theme:change:ace', this.handleAceThemeChangeEvent.bind(this));
+        $('#start_chat').on('click', function(event) {
+            this.createEventHandler('pp_start_chat', null)(event)
+            // Allow to open the new tab even in Safari.
+            // See: https://stackoverflow.com/a/70463940
+            setTimeout(() => {
+                var pop_up_window = window.open($('#start_chat').data('url'), '_blank');
+                if (pop_up_window) {
+                    pop_up_window.onerror = function (message) {
+                        $.flash.danger({text: message});
+                        this.sendError(message, null);
+                    };
+                }
+            })
+        }.bind(this));
         this.initializeFileTreeButtons();
         this.initializeWorkspaceButtons();
         this.initializeRequestForComments()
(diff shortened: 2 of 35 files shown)


app/views/exercises/implement.html.slim
```diff
@@ -7,6 +7,10 @@
   - unless @embed_options[:hide_exercise_description]
     .exercise.clearfix
       div
+        - if current_contributor.programming_group?
+          a.btn.btn-primary(href='https://jitsi.fem.tu-ilmenau.de/openHPI_ProgrammingGroup#{current_contributor.id}', target='_blank')
```
MrSerth, line 11: We should probably add some data-privacy statement here...

app/controllers/sessions_controller.rb
```diff
@@ -22,7 +22,7 @@ def create_through_lti
     store_nonce(params[:oauth_nonce])
     if params[:custom_redirect_target]
       redirect_to(URI.parse(params[:custom_redirect_target].to_s).path)
-    elsif PairProgramming23Study.participate?
+    elsif PairProgramming23Study.participate?(current_user, @exercise, session[:study_group_id])
```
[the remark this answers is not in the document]
kiragrammel: I just tested it and `current_user.current_study_group_id` is in this function `nil`.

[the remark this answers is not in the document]
MrSerth: I see. Then, we should fix [this method](https://github.com/openHPI/codeocean/blob/ec9be819428f5c93f32e1f45c6facfb6a92e5b11/app/controllers/concerns/lti.rb#L172) by calling the `‎store_current_study_group_id` method as performed in the [`ApplicationController`](https://github.com/openHPI/codeocean/blob/ec9be819428f5c93f32e1f45c6facfb6a92e5b11/app/controllers/application_controller.rb#L24).

config/locales/de.yml
```diff
@@ -466,6 +468,7 @@ de:
       default_linter_feedback: Sehr gut. Der Linter hat nichts mehr zu beanstanden.
       error_messages: Fehlermeldungen
       existing_programming_group: Sie arbeiten gerade an der Übung mit dem Titel '%{exercise}' als Teil einer Programmiergruppe. Bitte schließen Sie Ihre Arbeit dort ab, indem Sie Ihren Code bewerten und abgeben, bevor Sie mit der Bearbeitung dieser Übung beginnen.
+      external_data_policy: Durch Nutzung des Videochats stimmen Sie der <a href=%{url}>externen Datenschutzerklärung</a> zu.
```
kiragrammel, line 471: The link is not clickable.

MrSerth: You might need to use two == in the view.

MrSerth:
Thanks for these changes! Since there are currently no open comments on this PR left, I deployed the current version to Staging and CodeMoon for further testing 🧪.

app/views/exercises/implement.html.slim
```diff
@@ -6,12 +6,19 @@
 #editor-column
   - unless @embed_options[:hide_exercise_description]
     .exercise.clearfix
-      div
-        span.badge.rounded-pill.bg-primary.float-end.score
+      div.col-3.float-end
```
MrSerth, line 9: We still have some issues with smaller screens here, and could potentially improve the layout a little. Not too important, but just something I noticed.

Bildschirmfoto 2023-09-03 um 23 03 52

app/assets/javascripts/channels/synchronized_editor_channel.js
```diff
@@ -24,12 +24,12 @@ $(document).on('turbolinks:load', function () {
         received(data) {
           // Called when there's incoming data on the websocket for this channel
           if (current_user_id !== data['current_user_id']) {
```
MrSerth, line 26: This condition prevents a single user from opening the same exercise in two windows. Further, we should not just use the user ID, since the user type needs to match as well.

Potentially, we could also switch to a random identifier (such as a UUID for each page load), and use that one as identifier. It should resolve those issues, I assume.

app/assets/javascripts/editor/editor.js.erb
```diff
@@ -339,7 +339,7 @@ var CodeOceanEditor = {
                     CodeOceanEditor.lastDeltaObject = null;
```
MrSerth, line 339: I am currently not sure whether we should change the `lastDeltaObject` as well. In theory, I imagine that a race condition could occur: The `lastDeltaObject` is "global", in a sense that it is used for all open editors (i.e., all files). Potentially, there could be the same change in two distinct files. If, and that is very unlikely, the second change and event handler is faster than the assignment of `CodeOceanEditor.lastDeltaObject = null;`, we would ignore the second change (and, for example, not send it through the WebSocket connection).

Practically, I don't think that this could happen (due to human timing), but just wanted to point it out.

MrSerth (commented):
I tested the latest changes and identified a few minor things to consider. But luckily, this PR will close #1868, which is awesome! 🎉

config/locales/de.yml
```diff
@@ -57,7 +57,7 @@ de:
         uuid: UUID
         unpublished: Deaktiviert
       programming_group:
-        programming_partner_ids: Nutzer-IDs der Programmierpartner
+        programming_partner_id: Nutzer-ID der Programmierpartnerin/ des Programmierpartners
```
MrSerth, line 60:
```suggestion
        programming_partner_id: Nutzer-ID der Programmierpartnerin / des Programmierpartners
```

config/locales/en.yml
```diff
@@ -466,6 +471,7 @@ en:
       default_linter_feedback: Well done. The linter is completly satisfied.
       error_messages: Error Messages
       existing_programming_group: You are currently working on the exercise entitled '%{exercise}' as part of a programming group. Please finish your work there by scoring and submitting your code before you start implementing this exercise.
+      external_data_policy: By using the video chat, you agree to the <a href=%{url}> external data policy</a>.
```
MrSerth, line 474:
```suggestion
      external_data_policy: By using the video chat, you agree to the <a href=%{url}>external privacy policy</a>.
```

app/assets/javascripts/channels/synchronized_editor_channel.js
```diff
@@ -23,13 +24,18 @@ $(document).on('turbolinks:load', function () {
 
         received(data) {
           // Called when there's incoming data on the websocket for this channel
-          if (current_user_id !== data['current_user_id']) {
-            CodeOceanEditor.applyChanges(data['delta']['data']);
+          if (data.command === 'editor_change' && current_user_id !== data['current_user_id']) {
```
MrSerth, line 27: As discussed, we should not just rely on the user ID here, but also consider the user type. Further, we should switch to some generic UUID where possible, to prevent some synchronization issues with multiple tabs / windows.

MrSerth, line 27: Since you are comparing the `current_user_id !== data['current_user_id']` in all conditions, this could be done first. Then, you could also use a switch statement (optional).

app/assets/javascripts/editor/editor.js.erb
```diff
@@ -1043,6 +1058,15 @@ var CodeOceanEditor = {
           delta_data.text === last_delta.text;
     },
 
+    showUserIsOnline: function (status_and_user) {
+        if (status_and_user.status === 'connected') {
+            $('#pg_session').text(status_and_user.current_user_name + I18n.t('exercises.editor.is_online'));
```
MrSerth, line 1063: What if... we change the translation, so that it already allows passing a user name?

```suggestion
            $('#pg_session').text(I18n.t('exercises.editor.is_online', {name: status_and_user.current_user_name}));
```

This, of course, also requires a corresponding change in the en.yml / de.yml.

app/assets/javascripts/editor/editor.js.erb
```diff
@@ -1043,6 +1058,15 @@ var CodeOceanEditor = {
           delta_data.text === last_delta.text;
     },
 
+    showUserIsOnline: function (status_and_user) {
```
MrSerth, line 1061: The name of this method seems somewhat misleading, since it could also show that a user is offline.

app/assets/javascripts/channels/synchronized_editor_channel.js
```diff
@@ -23,13 +24,18 @@ $(document).on('turbolinks:load', function () {
...
+              CodeOceanEditor.applyChanges(data['delta']['data'], data['active_file']);
+          } else if (data.command === 'connection_change' && current_user_id !== data.current_user_id) {
+              CodeOceanEditor.showUserIsOnline(data);
```
MrSerth, line 30: Rather than passing the data, what if we would only pass the name and online status? Then, the method would be more independent of the actual data passed here.

app/assets/javascripts/channels/synchronized_editor_channel.js
```diff
@@ -23,13 +24,18 @@ $(document).on('turbolinks:load', function () {
...
+              this.perform('send_hello', {command: 'hello', status: 'connected', current_user_id: current_user_id, current_user_name: current_user_name});
```
MrSerth, line 31: It feels like your `send_hello` method introduced in the backend is just a simple echo / forwarding, that would allow sending arbitrary data between the peers. While that works, for sure, it feels like most information is already present server-side, where we are also sure that it cannot be manipulated. Hence, I would add all this information just there...

app/views/exercises/_editor.html.slim
```diff
@@ -39,9 +39,22 @@
             = t('exercises.editor.download')
 
       div
+        ruby:
+          if current_contributor.programming_group?
+            current_contributor.users.each do |user|
+              if user.id != current_user.id
+                @programming_partners_name = user.name
```
MrSerth, line 46: If I am working with multiple partners, this variable is overwritten. Hence, only one of my partners is listed there.

app/views/exercises/implement.html.slim
```diff
@@ -6,16 +6,28 @@
...
+      h1 id="exercise-headline"
+        i id="description-symbol" class=(@embed_options[:collapse_exercise_description] ? 'fa-solid fa-chevron-right' : 'fa-solid fa-chevron-down')
+        = @exercise.title
 
       #description-card.lead class=(@embed_options[:collapse_exercise_description] ? 'description-card-collapsed' : 'description-card')
         = render_markdown(@exercise.description)
 
+      - if current_contributor.programming_group?
+        .small.text-body-tertiary
+          == t('exercises.implement.pair_programming_feedback', url:'https://etherpad.xopic.de/p/openHPI_PairProgrammingFeedback')
```
MrSerth, line 29:
```suggestion
          == t('exercises.implement.pair_programming_feedback', url: 'https://etherpad.xopic.de/p/openHPI_PairProgrammingFeedback')
```

Within the Etherpad, I would replace "Bugs" with something more generic (that is more easily understandable by our target group).

config/locales/de.yml
```diff
@@ -481,6 +487,7 @@ de:
       exit_failure: Die letzte Code-Ausführung wurde mit einem Fehler beendet (Statuscode %{exit_code}).
       no_output_yet: Bisher existiert noch keine Ausgabe.
       output: Programm-Ausgabe
+      pair_programming_feedback: Falls dir Fehler bei dem Pair Programming Feature auffallen, kannst du diese gerne <a target='_blank', href=%{url}>hier aufschreiben</a>. Vielen Dank!"
```
MrSerth, line 490: I was just thinking about making this a little more German ("Probleme bei der Zusammenarbeit...")

MrSerth (commented):
Except for the latest commit with the additional events, I've reviewed your changes already.

app/channels/synchronized_editor_channel.rb
```diff
@@ -3,11 +3,13 @@
 class SynchronizedEditorChannel < ApplicationCable::Channel
   def subscribed
     stream_from specific_channel
+    ActionCable.server.broadcast(specific_channel, {command: 'connection_change', status: 'connected', current_user_id: current_user.id, current_user_name: current_user.name})
```
MrSerth, line 6: Here, and also below, we should potentially use the user_type as well (if this change is still present after switching to a UUID.

MrSerth:
This PR has grown in size and brings many nice improvements already. Nevertheless, some aspects are still required to be worked on. 

Since we are requiring (some) changes right now, I'll proceed and merge them now. Nevertheless, we should keep track of the open issues and resolve them in a follow-up MR.

app/views/exercises/_editor.html.slim
MrSerth: As discussed, it would be nice to show the status for larger groups as well. Potentially, there are three cases:

- All others are offline
- Some are online
- All others are online

We could add a hover tooltip to show the names.

PR #1884, opened 2023-09-08
Layout improvements  & Info for pair programming

kiragrammel:
* Add info modal for pair programming
* Improve layout for programming groups

### DIFF
--- a/app/views/exercises/implement.html.slim
+++ b/app/views/exercises/implement.html.slim
@@ -6,25 +6,31 @@
 #editor-column
   - unless @embed_options[:hide_exercise_description]
     .exercise.clearfix
-      div.col-3.float-end
-        span.badge.rounded-pill.bg-primary.float-end.mt-2.score
-        - if current_contributor.programming_group?
-          button.btn.btn-sm.btn-primary.me-3.mt-1#start_chat  data= {url: "https://jitsi.fem.tu-ilmenau.de/openHPI_ProgrammingGroup#{current_contributor.id}"}
-            i.fa-solid.fa-video
+      div.col-9div.d-lg-flex.flex-row.justify-content-between.align-items-baseline
+        div.col-lg-7
+          h1 id="exercise-headline"
+            i id="description-symbol" class=(@embed_options[:collapse_exercise_description] ? 'fa-solid fa-chevron-right' : 'fa-solid fa-chevron-down')
+            = @exercise.title
 
-            span = t('exercises.editor.start_video')
+        div.col-lg-5.float-lg-end.ms-md-3.mb-md-3.d-flex.justify-content-end.pe-lg-3
+          - if current_contributor.programming_group?
+            span.me-3.mt-1
+              button.btn.btn-sm.btn-primary.text-nowrap.w-100#start_chat data= { url: "https://jitsi.fem.tu-ilmenau.de/openHPI_ProgrammingGroup#{current_contributor.id}" }
+                i.fa-solid.fa-video
+                = t('exercises.editor.start_video')
 
-          div.small.text-body-tertiary.mt-1
-            == t('exercises.implement.external_privacy_policy', url:'https://www.tu-ilmenau.de/datenschutz')
+              div.small.text-body-tertiary.text-nowrap.mt-1
+                == t('exercises.implement.external_privacy_policy', url: 'https://www.tu-ilmenau.de/datenschutz')
 
-          button.btn.btn-sm.btn-secondary.me-3.mt-1 data-bs-toggle= 'modal' data-bs-target= '#modal-info-pair-programming'
-            i.fa-solid.fa-circle-info
-            = t('programming_groups.new.pair_programming_info')
+            span.me-3.mt-1
+              button.btn.btn-sm.btn-secondary.text-nowrap#info data-bs-toggle= 'modal' data-bs-target= '#modal-info-pair-programming'
+                i.fa-solid.fa-circle-info
+                = t('programming_groups.new.pair_programming_info')
 
+            span.badge.rounded-pill.bg-primary.float-end.mt-2.mb-5.score
 
-      h1 id="exercise-headline"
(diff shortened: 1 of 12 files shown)


app/views/programming_groups/new.html.slim
```diff
@@ -5,18 +5,22 @@ h1 = t('programming_groups.new.create_programming_pair')
...
-        == t('programming_groups.new.work_alone', path: implement_exercise_path(@exercise))
+      == t('programming_groups.new.work_alone', path: implement_exercise_path(@exercise))
 
   .col-md-6
     h5 = t('programming_groups.new.find_partner_title')
     p
       = t('programming_groups.new.find_partner_description')
     iframe name="embed_readwrite" src="https://etherpad.xopic.de/p/find_programming_group_for_exercise_#{@exercise.id}?showControls=true&showChat=true&showLineNumbers=true&useMonospaceFont=false" width="100%" height="300" style='border: 1px solid black;'
+
+= render('shared/modal', classes: 'modal-lg', id: 'modal-info-pair-programming', template: 'application/info_pair_programming', title: "Pair Programming Info")
```
MrSerth, line 26: Potentially, you could / should use the translation `shared.pair_programming_info`:

```suggestion
= render('shared/modal', classes: 'modal-lg', id: 'modal-info-pair-programming', template: 'application/info_pair_programming', title: t('shared.pair_programming_info'))
```

app/views/exercises/implement.html.slim
```diff
@@ -36,3 +47,4 @@
 
 
   = render('editor', exercise: @exercise, files: @files, submission: @submission)
+  = render('shared/modal', classes: 'modal-lg', id: 'modal-info-pair-programming', template: 'application/info_pair_programming', title: "Pair Programming Info")
```
MrSerth, line 50: Potentially, you could / should use the translation `shared.pair_programming_info`:

```suggestion
  = render('shared/modal', classes: 'modal-lg', id: 'modal-info-pair-programming', template: 'application/info_pair_programming', title: t('shared.pair_programming_info'))
```

app/views/application/info_pair_programming.html.slim
```diff
@@ -0,0 +1,7 @@
+= render_markdown(t('programming_groups.new.info_pair_programming'))
+= render_markdown(t('programming_groups.new.info_study'))
+== t('exercises.implement.pair_programming_feedback', url: 'https://etherpad.xopic.de/p/openHPI_PairProgrammingFeedback')
+
+- content_for :modal_footer do
+  button.btn.btn-primary data-bs-dismiss= 'modal' onclick= 'CodeOceanEditor.setStoredViewedPPInfo(true);'
```
MrSerth, line 6: Some spacing suggestions:

```suggestion
  button.btn.btn-primary data-bs-dismiss='modal' onclick='CodeOceanEditor.setStoredViewedPPInfo(true);'
```

MrSerth, line 6: For me, the `onclick` handler is not working (which is inline with my expectation) due to the [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP):

`
Refused to execute inline event handler because it violates the following Content Security Policy directive: "script-src 'self' 'report-sample'". Either the 'unsafe-inline' keyword, a hash ('sha256-...'), or a nonce ('nonce-...') is required to enable inline execution. Note that hashes do not apply to event handlers, style attributes and javascript: navigations unless the 'unsafe-hashes' keyword is present.
`

Since we won't change the CSP, you'll need to find another solution.

app/views/application/info_pair_programming.html.slim
MrSerth, line 1: Since this file is a partial, it should start with an underscore in it's file name.

MrSerth, line 1: Putting this file into the application folder doesn't fit right. I was shortly considering putting it into the `shared` folder (which would be better), but we could also put it into the programming groups view. Then, from the exercise view, you should also be able to use it.

app/views/application/info_pair_programming.html.slim
```diff
@@ -0,0 +1,7 @@
...
+    = t('programming_groups.new.dont_show_modal_again')
```
MrSerth, line 7: My proposal would be to have two buttons on this modal: One to close it (blue, primary), and another one to hide it on this device (potentially just gray, secondary). This way, learners might not close it unintended...

app/assets/javascripts/editor/editor.js.erb
```diff
@@ -1083,6 +1092,10 @@ var CodeOceanEditor = {
         this.initializeDeadlines();
         CodeOceanEditorTips.initializeEventHandlers();
 
+        if (this.getStoredViewedPPInfo() == true) {
```
MrSerth, line 1095: We should also do a comparison with three `===` signs. I assume, you chose this comparison because the return value will be `'true'` (as string) rather than `true` (as boolean). That's fine, just keep the string representation.

app/assets/javascripts/editor/editor.js.erb
```diff
@@ -30,6 +30,7 @@ var CodeOceanEditor = {
 
     lastCopyText: null,
     lastDeltaObject_for_file: new Map(),
+    viewed_pp_info: false,
```
MrSerth, line 33: I am not quite sure why you need this variable, since it is not used.

app/assets/javascripts/editor/editor.js.erb
```diff
@@ -1083,6 +1092,10 @@ var CodeOceanEditor = {
         this.initializeDeadlines();
         CodeOceanEditorTips.initializeEventHandlers();
 
+        if (this.getStoredViewedPPInfo() == true) {
+            new bootstrap.Modal($('#modal-info-pair-programming')).show();
```
MrSerth, line 1096: For me, the modal doesn't show automatically yet...

MrSerth (commented):
Here's a short review on your changes. I am really happy about the layout changes we introduced, and think they look pretty nice (at least for our frontend skills 😜)

PR #1892, opened 2023-09-12
Refactor pair programming & solve problems with synchronization

kiragrammel:
* Replaces compareDeltaObjects with editor.curOp
* Reload text for person joining pp session
* Generate Session ID on server for synchronized editor
* Synchronize resetCode for pair programming


MrSerth (commented):
Awesome, we're getting there! 👏

MrSerth (approved):

MrSerth:
The error we see here, `ERROR: payload string too long` is already tracked in #1920 and might be resolved with switching to another ActionCable adapter.

PR #1893, opened 2023-09-12
Correct errors in pair programming

kiragrammel:


MrSerth:
I've fixed a minor issue with the display name and missing string interpolation for Safari users.

MrSerth (approved):

PR #1906, opened 2023-09-14
Create index, destroy, show, edit and update route for programming groups

### LINKED ISSUE #1848: Add show route for programming_group
For the submissions index view, we only implemented a quick fix for the programming groups so that no links are shown but only the display name.
To fix it completely, the show route for the programming groups needs to be implemented.

_Originally posted by @MrSerth in https://github.com/openHPI/codeocean/pull/1830#discussion_r1292856503_

kiragrammel:
* Add an index route for programming groups to have an overview for admins which groups exist
* Add an destroy route for programming groups to delete groups and make it possible to start working with another person on a exercise or work alone
* Add edit & update route for programming groups to remove users from a group
* closes #1848

### DIFF
--- a/app/controllers/programming_groups_controller.rb
+++ b/app/controllers/programming_groups_controller.rb
@@ -4,7 +4,17 @@ class ProgrammingGroupsController < ApplicationController
   include CommonBehavior
   include LtiHelper
 
-  before_action :set_exercise_and_authorize
+  before_action :set_exercise_and_authorize, only: %i[new create]
+  before_action :set_programming_group_and_authorize, only: MEMBER_ACTIONS
+
+  def index
+    set_exercise_and_authorize if params[:exercise_id].present?
+    @search = ProgrammingGroup.ransack(params[:q], {auth_object: current_user})
+    @programming_groups = @search.result.includes(:exercise, :programming_group_memberships, :internal_users, :external_users).order(:id).paginate(page: params[:page], per_page: per_page_param)
+    authorize!
+  end
+
+  def show; end
 
   def new
     Event.create(category: 'page_visit', user: current_user, exercise: @exercise, data: 'programming_groups_new', file_id: nil)
@@ -23,9 +33,13 @@ class ProgrammingGroupsController < ApplicationController
     end
   end
 
+  def edit
+    @members = @programming_group.programming_group_memberships.includes(:user)
+  end
+
   def create
-    programming_partner_ids = programming_group_params[:programming_partner_ids].split(',').map(&:strip).uniq
-    users = programming_partner_ids.map do |partner_id|
+    programming_partner_ids = programming_group_params&.fetch(:programming_partner_ids, [])&.split(',')&.map(&:strip)&.uniq
+    users = programming_partner_ids&.map do |partner_id|
       User.find_by_id_with_type(partner_id)
     rescue ActiveRecord::RecordNotFound
       partner_id
@@ -33,12 +47,12 @@ class ProgrammingGroupsController < ApplicationController
     @programming_group = ProgrammingGroup.new(exercise: @exercise, users:)
     authorize!
 
-    unless programming_partner_ids.include? current_user.id_with_type
+    unless programming_partner_ids&.include? current_user.id_with_type
       @programming_group.add(current_user)
     end
 
     unless @programming_group.valid?
-      Event.create(category: 'pp_invalid_partners', user: current_user, exercise: @exercise, data: programming_group_params[:programming_partner_ids], file_id: nil)
+      Event.create(category: 'pp_invalid_partners', user: current_user, exercise: @exercise, data: programming_group_params&.fetch(:programming_partner_ids), file_id: nil)
     end
 
     create_and_respond(object: @programming_group, path: proc { implement_exercise_path(@exercise) }) do
@@ -65,19 +79,36 @@ class ProgrammingGroupsController < ApplicationController
     end
   end
 
+  def update
+    myparams = programming_group_params || {}
+    @members = @programming_group.programming_group_memberships.includes(:user)
+    myparams[:users] = @members.where(id: myparams&.fetch(:programming_group_membership_ids, [])&.compact_blank).map(&:user)
+    update_and_respond(object: @programming_group, params: myparams)
+  end
+
+  def destroy
+    session.delete(:pg_id) if current_contributor == @programming_group
+    destroy_and_respond(object: @programming_group)
+  end
+
   private
 
   def authorize!
-    authorize(@programming_group)
+    authorize(@programming_group || @programming_groups)
   end
 
   def programming_group_params
-    params.require(:programming_group).permit(:programming_partner_ids)
+    params.require(:programming_group).permit(:programming_partner_ids, programming_group_membership_ids: []) if params[:programming_group].present?
   end
 
   def set_exercise_and_authorize
     @exercise = Exercise.find(params[:exercise_id])
     authorize(@exercise, :implement?)
+  end
+
+  def set_programming_group_and_authorize
+    @programming_group = ProgrammingGroup.find(params[:id])
+    authorize!
   end
 
   def redirect_to_exercise
--- a/app/models/exercise.rb
+++ b/app/models/exercise.rb
@@ -36,6 +36,7 @@ class Exercise < ApplicationRecord
   has_many :request_for_comments
 
   scope :with_submissions, -> { where('id IN (SELECT exercise_id FROM submissions)') }
+  scope :with_programming_groups, -> { where('id IN (SELECT exercise_id FROM programming_groups)') }
 
   validate :valid_main_file?
   validate :valid_submission_deadlines?
@@ -612,7 +613,7 @@ class Exercise < ApplicationRecord
   end
 
   def self.ransackable_attributes(_auth_object = nil)
-    %w[title internal_title]
+    %w[title id internal_title]
   end
 
   def self.ransackable_associations(_auth_object = nil)
--- a/app/models/programming_group.rb
+++ b/app/models/programming_group.rb
@@ -69,13 +69,14 @@ class ProgrammingGroup < ApplicationRecord
   end
 
   def users=(users)
-    self.internal_users = []
-    self.external_users = []
     users&.each do |user|
       next erroneous_users << user unless user.is_a?(User)
 
       add(user)
     end
+
+    # Remove all users that are no longer part of the programming group.
+    programming_group_memberships.where.not(user: users).destroy_all
   end
 
   def self.ransackable_associations(_auth_object = nil)
--- a/app/models/programming_group.rb
+++ b/app/models/programming_group.rb
@@ -28,8 +28,8 @@ class ProgrammingGroup < ApplicationRecord
     false
   end
 
-  def self.nested_resource?
-    true
+  def self.parent_resource
+    Exercise
   end
 
   def programming_group?
--- a/app/models/programming_group.rb
+++ b/app/models/programming_group.rb
@@ -9,21 +9,16 @@ class ProgrammingGroup < ApplicationRecord
   has_many :internal_users, through: :programming_group_memberships, source_type: 'InternalUser', source: :user
   has_many :testruns, through: :submissions
   has_many :runners, as: :contributor, dependent: :destroy
-  has_many :events
-  has_many :events_synchronized_editor, class_name: 'Event::SynchronizedEditor'
-  has_many :pair_programming_exercise_feedbacks
-  has_many :pair_programming_waiting_users
+  has_many :events, dependent: :destroy
+  has_many :events_synchronized_editor, class_name: 'Event::SynchronizedEditor', dependent: :destroy
+  has_many :pair_programming_exercise_feedbacks, dependent: :destroy
+  has_many :pair_programming_waiting_users, dependent: :destroy
   belongs_to :exercise
 
   validate :min_group_size
   validate :max_group_size
   validate :no_erroneous_users
   accepts_nested_attributes_for :programming_group_memberships
-
-  def initialize(attributes = nil)
-    @erroneous_users = []
-    super
-  end
 
   def external_user?
     false
@@ -76,14 +71,26 @@ class ProgrammingGroup < ApplicationRecord
   def users=(users)
     self.internal_users = []
     self.external_users = []
-    users.each do |user|
-      next @erroneous_users << user unless user.is_a?(User)
+    users&.each do |user|
+      next erroneous_users << user unless user.is_a?(User)
 
       add(user)
     end
   end
 
+  def self.ransackable_associations(_auth_object = nil)
+    %w[exercise programming_group_memberships]
+  end
+
+  def self.ransortable_attributes(_auth_object = nil)
+    %w[id created_at]
+  end
+
   private
+
+  def erroneous_users
+    @erroneous_users ||= []
+  end
 
   def min_group_size
     if users.size < 2
@@ -98,7 +105,7 @@ class ProgrammingGroup < ApplicationRecord
   end
 
   def no_erroneous_users
-    @erroneous_users.each do |partner_id|
+    erroneous_users.each do |partner_id|
       errors.add(:base, :invalid_partner_id, partner_id:)
     end
   end
--- a/app/views/programming_groups/index.html.slim
+++ b/app/views/programming_groups/index.html.slim
@@ -1,0 +1,43 @@
+- if params[:exercise_id].nil?
+  h1 = ProgrammingGroup.model_name.human(count: 2)
+  = render(layout: 'shared/form_filters') do |f|
+    .col-auto
+      = f.label(:exercise_id_eq, t('activerecord.attributes.programming_group.exercise'), class: 'visually-hidden form-label')
+      = f.collection_select(:exercise_id_eq, Exercise.with_programming_groups, :id, :title, class: 'form-control', prompt: t('activerecord.attributes.programming_group.exercise'))
+    .col-auto
+      = f.label(:programming_group_memberships_user_of_ExternalUser_type_id_eq, t('activerecord.attributes.programming_group.external_user_id'), class: 'visually-hidden form-label')
+      = f.search_field(:programming_group_memberships_user_of_ExternalUser_type_id_eq, class: 'form-control', placeholder: t('activerecord.attributes.programming_group.external_user_id'))
+    .col-auto
+      = f.label(:programming_group_memberships_user_of_InternalUser_type_id_eq, t('activerecord.attributes.programming_group.internal_user_id'), class: 'visually-hidden form-label')
+      = f.search_field(:programming_group_memberships_user_of_InternalUser_type_id_eq, class: 'form-control', placeholder: t('activerecord.attributes.programming_group.internal_user_id'))
+- else
+  h1 = "#{ProgrammingGroup.model_name.human(count: 2)} for Exercise '#{@exercise.title}'"
+
+.table-responsive
+  table.table.mt-4 class="#{@programming_groups.present? ? 'sortable' : ''}"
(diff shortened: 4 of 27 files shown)


app/views/programming_groups/index.html.slim
```diff
@@ -1,5 +1,10 @@
 h1 = ProgrammingGroup.model_name.human(count: 2)
 
+= render(layout: 'shared/form_filters') do |f|
```
kiragrammel, line 3: The filter is not filtering the table.

app/views/programming_groups/index.html.slim
```diff
@@ -0,0 +1,23 @@
...
+  .col-auto
+    = f.label(:exercise_id, t('activerecord.attributes.programming_group.exercise_title'), class: 'visually-hidden form-label')
+    = f.collection_select(:exercise_id, Exercise.with_programming_groups, :id, :title, class: 'form-control', prompt: t('activerecord.attributes.programming_group.exercise_title'))
+
+.table-responsive
+  table.table.mt-4
+    thead
+      tr
+        th = t('activerecord.attributes.programming_group.programming_group_id')
```
MrSerth, line 12: This is not the ID but rather the name, isn't it?

app/views/programming_groups/index.html.slim
```diff
@@ -0,0 +1,23 @@
...
+        th = t('activerecord.attributes.programming_group.exercise_title')
```
MrSerth, line 13: Let's rather go with `Exercise`  than `Exercise title`

app/controllers/programming_groups_controller.rb
```diff
@@ -4,7 +4,14 @@ class ProgrammingGroupsController < ApplicationController
   include CommonBehavior
   include LtiHelper
 
-  before_action :set_exercise_and_authorize
+  before_action :set_exercise_and_authorize, only: %i[new create]
+  before_action :set_programming_group, only: %i[destroy]
+
+  def index
+    @search = ProgrammingGroup.ransack(params[:q], {auth_object: current_user})
+    @programming_groups = @search.result.includes(:exercise).order(:id).paginate(page: params[:page], per_page: per_page_param)
```
MrSerth, line 12: This includes is not sufficient, we are still performing an N+1 query for the programming group memberships and users...

Let's go with:

```rb
includes(:exercise, :programming_group_memberships, :internal_users, :external_users)
```

Please take a moment to compare the SQL queries generated by your code and my suggestion to see the difference 🔍

app/views/programming_groups/index.html.slim
```diff
@@ -0,0 +1,23 @@
...
+        th = t('activerecord.attributes.programming_group.participants')
+        th = t('shared.actions')
+    tbody
+      - @programming_groups.each do |programming_group|
+        tr
+          td = programming_group.displayname
+          td = link_to_if(policy(programming_group.exercise).show?, programming_group.exercise.title, programming_group.exercise, 'data-turbolinks' => "false")
+          td == programming_group.users.map { |user| link_to_if(policy(user).show?, user.name, user) }.join(', ')
```
MrSerth, line 21: For groups with many users, this might be too much for the column. For these cases, we might think of some good truncation (e.g., via CSS). For now, it might be fine.

config/routes.rb
```diff
@@ -100,6 +100,8 @@
     resources :programming_groups, only: %i[new create]
```
MrSerth, line 100: We might also think to link the programming groups here, in which case they should be pre-filtered for the given exercise. You can determine whether we are in a scoped environment with the `params[:exercise_id]` (and, if called, the `@exercise` variable). In that case, the Ransack part (in the controller and the view) can be hidden.

app/views/programming_groups/index.html.slim
MrSerth, line 1: When visiting this page, the `PROGRAMMING GROUPS` in the breadcrumb navigation is not clickable. However, it is for exercises and most other pages...

app/views/programming_groups/index.html.slim
```diff
@@ -0,0 +1,23 @@
+h1 = ProgrammingGroup.model_name.human(count: 2)
+
+= render(layout: 'shared/form_filters') do |f|
+  .col-auto
+    = f.label(:exercise_id, t('activerecord.attributes.programming_group.exercise_title'), class: 'visually-hidden form-label')
+    = f.collection_select(:exercise_id, Exercise.with_programming_groups, :id, :title, class: 'form-control', prompt: t('activerecord.attributes.programming_group.exercise_title'))
+
+.table-responsive
+  table.table.mt-4
```
MrSerth, line 9: Feel free to make the table sortable, by adding:

```
class="#{@programming_groups.present? ? 'sortable' : ''}"
```

Please ensure that sorting in both directions works as expected (will be sorted in the frontend by the client).

app/views/programming_groups/index.html.slim
MrSerth, line 12: You can make this (and potentially other columns) sortable in the backend by adding:

```rb
th = sort_link(@search, :id, t('activerecord.attributes.programming_group.programming_group_id'))
```

This might be useful to always have the newest entry first in the list.

app/views/programming_groups/index.html.slim
```diff
@@ -0,0 +1,23 @@
...
+    thead
+      tr
+        th = t('activerecord.attributes.programming_group.programming_group_id')
+        th = t('activerecord.attributes.programming_group.exercise_title')
+        th = t('activerecord.attributes.programming_group.participants')
+        th = t('shared.actions')
```
MrSerth, line 15: Other lists also feature some other nice columns. Two interesting ones I thought of might be the `created_at` or just a column with the number of users.

MrSerth (commented):
Nice, I am looking forward to bringing this nice view live! Here are just a few remarks:

Based on Study Group show and edit:
- programming_groups/:id/show
- programming_groups/:id/edit

Based on ExternalUser exercise statistics (might be another ticket):
- programming_groups/:id/statistics

app/models/exercise.rb
```diff
@@ -35,6 +35,7 @@ class Exercise < ApplicationRecord
   has_many :pair_programming_waiting_users
 
   scope :with_submissions, -> { where('id IN (SELECT exercise_id FROM submissions)') }
+  scope :with_programming_groups, -> { where('id IN (SELECT DISTINCT exercise_id FROM programming_groups)') }
```
MrSerth, line 38: I would argue that the `DISTINCT` operator here is not necessary and rather adds unrequired overhead. In this case, I would suggest dropping it.

app/views/programming_groups/index.html.slim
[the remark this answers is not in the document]
MrSerth: Let's go with `Contributors > Programming Groups`.

MrSerth:
Add a link to the respective programming group list for a given exercise (on exercise#index and exercise#show) in the dropdown.

app/views/programming_groups/index.html.slim
kiragrammel: This is only working for the exercise_id. Maybe do I need to add for each column a filter so that it is sortable?

app/views/programming_groups/index.html.slim
kiragrammel: The sorting is only working in one direction (asc) when clicking on it. In the url it is possible to change it.

app/controllers/programming_groups_controller.rb
```diff
@@ -65,21 +81,42 @@ def create
     end
   end
 
+  def update
```
kiragrammel, line 84: Not working yet

MrSerth (commented):
Here are some pending comments I already made last week (but didn't submit).

MrSerth:
> Add a link to the respective programming group list for a given exercise (on exercise#index and exercise#show) in the dropdown.

- [ ] For the index route, I would suggest putting the link right after the RfCs. 
- [ ] In the show route, a link to the programming groups is missing.

app/views/programming_groups/index.html.slim
```diff
@@ -0,0 +1,34 @@
...
+      = f.collection_select(:exercise_id_eq, Exercise.with_programming_groups, :id, :title, class: 'form-control', prompt: t('activerecord.attributes.programming_group.exercise'))
+
+.table-responsive
+  table.table.mt-4 class="#{@programming_groups.present? ? 'sortable' : ''}"
+    thead
+      tr
+        th = sort_link(@search, :id, t('activerecord.attributes.programming_group.programming_group_name'))
+        th = sort_link(@search, :exercise_id, t('activerecord.attributes.programming_group.exercise'))
+        th = sort_link(@search, :user_id, t('activerecord.attributes.programming_group.member'))
+        th = sort_link(@search, :count, t('activerecord.attributes.programming_group.member_count'))
```
MrSerth, line 16: Same as above: Sorting in the backend is not that trivial, even though possible. Hence, I would probably skip it for now, this might be some feature in the future. If desired, we could add a filter (e.g., a dropdown to see those with exactly 2 members and those with more), but since we currently limited all groups to two, I wouldn't add this filter either).

app/views/programming_groups/index.html.slim
```diff
@@ -0,0 +1,34 @@
...
+        th colspan=3 = t('shared.actions')
+    tbody
+      - if params[:exercise_id].nil?
+        - filtered_programming_groups = @programming_groups
+      - else
+        - filtered_programming_groups = @programming_groups.where(exercise_id: params[:exercise_id])
+      - filtered_programming_groups.each do |programming_group|
+        tr
+          td = link_to_if(policy(programming_group).show?, programming_group.displayname, programming_group)
+          td = link_to_if(policy(programming_group.exercise).show?, programming_group.exercise.title, programming_group.exercise, 'data-turbolinks' => "false")
```
MrSerth, line 27: Since we might render this page in the context of a specific exercise (`params[:exercise_id]` is set), we should remove this column there as well.

app/controllers/programming_groups_controller.rb
```diff
@@ -4,7 +4,18 @@ class ProgrammingGroupsController < ApplicationController
   include CommonBehavior
   include LtiHelper
 
-  before_action :set_exercise_and_authorize
+  before_action :set_exercise_and_authorize, only: %i[new create]
+  before_action :set_programming_group, only: %i[show edit update destroy]
```
MrSerth, line 8: We have a variable for that:

```suggestion
  before_action :set_programming_group, only: MEMBER_ACTIONS
```

app/controllers/programming_groups_controller.rb
```diff
@@ -65,21 +81,42 @@ def create
...
+  end
+
   def set_exercise_and_authorize
     @exercise = Exercise.find(params[:exercise_id])
     authorize(@exercise, :implement?)
   end
 
+  def set_programming_group
+    @programming_group = ProgrammingGroup.find(params[:id])
+  end
```
MrSerth, line 118: Let's ensure we are always authorizing:

```suggestion
  def set_programming_group_and_authorize
    @programming_group = ProgrammingGroup.find(params[:id])
    authorize!
  end
```

Please remove the "manual" `authorize!` from all `MEMBER_ACTIONS`.

app/models/programming_group.rb
MrSerth, line 1: You'll need to change the override of `initialize` to:

```rb
  after_initialize :initialize_erroneous_users

  def initialize_erroneous_users
    @erroneous_users = []
  end
```

app/controllers/programming_groups_controller.rb
```diff
@@ -65,21 +81,42 @@ def create
     end
   end
 
+  def update
+    myparams = programming_group_params2
+    myparams[:users] =
+      @programming_group.programming_group_memberships.where(id: myparams[:programming_group_membership_ids].compact_blank).includes(:user).map(&:user)
+    myparams.delete(:programming_group_membership_ids)
+    update_and_respond(object: @programming_group, params: myparams)
```
MrSerth, line 89: In case the update is not successful, we need to set the members as well:

```rb
    @members = @programming_group.programming_group_memberships.includes(:user)
```

app/controllers/programming_groups_controller.rb
```diff
@@ -65,21 +81,42 @@ def create
...
   def authorize!
-    authorize(@programming_group)
+    authorize(@programming_group || @programming_groups)
   end
 
   def programming_group_params
     params.require(:programming_group).permit(:programming_partner_ids)
   end
 
+  def programming_group_params2
```
MrSerth, line 107: A method name with a `2` suffix is not that great and recognizable. Please describe what they are intended for, e.g., `programming_group_params_for_update`.

MrSerth (commented):
I think, these are the main, missing points. Nothing too big, only a few remarks. I am sure, we can proceed merging soon ™️

app/controllers/programming_groups_controller.rb
MrSerth: For the StudyGroupsController#update, I fixed that as well. Potentially, you could update your code to be similar: https://github.com/openHPI/codeocean/commit/052367c1861028df5dcbe174da664175b7deee8f

app/views/programming_groups/index.html.slim
kiragrammel: I mentioned it in issue #1961.

app/views/programming_groups/index.html.slim
kiragrammel: With your comments below it is working now. 🙂

app/views/programming_groups/index.html.slim
```diff
@@ -0,0 +1,34 @@
...
+      = f.label(:exercise_id_eq, t('activerecord.attributes.programming_group.exercise'), class: 'visually-hidden form-label')
+      = f.collection_select(:exercise_id_eq, Exercise.with_programming_groups, :id, :title, class: 'form-control', prompt: t('activerecord.attributes.programming_group.exercise'))
+
+.table-responsive
+  table.table.mt-4 class="#{@programming_groups.present? ? 'sortable' : ''}"
+    thead
+      tr
+        th = sort_link(@search, :id, t('activerecord.attributes.programming_group.programming_group_name'))
+        th = sort_link(@search, :exercise_id, t('activerecord.attributes.programming_group.exercise'))
+        th = sort_link(@search, :user_id, t('activerecord.attributes.programming_group.member'))
```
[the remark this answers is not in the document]
kiragrammel: I tried to add a filter for the user ID with adding the following to code snippets to the index.html.slim:
```
   .col-auto
      = f.label(:external_user_id_cont, t('activerecord.attributes.programming_group.member'), class: 'visually-hidden form-label')
      = f.search_field(:external_user_id_cont, class: 'form-control', placeholder: t('activerecord.attributes.programming_group.member'))
```

and
```
th.sorttable_nosort = sort_link(@search, :external_user_id, t('activerecord.attributes.programming_group.member'))
```

And adding `external_user` to the ransackable_associations. But I get the following error:
```
undefined method `external_user_id_cont' for Ransack::Search<class: ProgrammingGroup, base: Grouping <combinator: and>>:Ransack::Search
```

app/views/programming_groups/index.html.slim
```diff
@@ -0,0 +1,34 @@
...
+  = render(layout: 'shared/form_filters') do |f|
+    .col-auto
+      = f.label(:exercise_id_eq, t('activerecord.attributes.programming_group.exercise'), class: 'visually-hidden form-label')
+      = f.collection_select(:exercise_id_eq, Exercise.with_programming_groups, :id, :title, class: 'form-control', prompt: t('activerecord.attributes.programming_group.exercise'))
+
+.table-responsive
+  table.table.mt-4 class="#{@programming_groups.present? ? 'sortable' : ''}"
+    thead
+      tr
+        th = sort_link(@search, :id, t('activerecord.attributes.programming_group.programming_group_name'))
```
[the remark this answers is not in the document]
MrSerth: Adding `sorttable_nosort` only makes sense when sorting in the frontend is allowed with `class="#{@programming_groups.present? ? 'sortable' : ''}"` (which got removed recently).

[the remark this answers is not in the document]
kiragrammel: I have added it again.

app/views/programming_groups/index.html.slim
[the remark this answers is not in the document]
MrSerth: Alright, I see. Indeed, your changes look mostly fine, there is just a minor thing left with the configuration of Ransack (the permitted attributes / associations) that is preventing the search to work correctly.

If I am not mistaken, the following column is not required, is it? In that case (and because it would still have the same error as I mentioned before), I would just omit it:

```slim
th.sorttable_nosort = sort_link(@search, :external_user_id, t('activerecord.attributes.programming_group.member'))
```

Besides that, let's dig into further details of the `external_user_id_cont`:

1. The `_cont` ending is not ideal. Since we are dealing with an ID column, what does `cont` mean here semantically? Is `4` contained in the ID `42`? As usual, I would switch to `eq` for ID columns.
2. You are trying to access the `external_user` information here. This leaves out all internal users. Further, no direct access is available (but rather only through the programming group memberships). Hence, we always need to prefix with `programming_group_memberships_`. There are two solutions I can think of, which are outlined in the following.

- Either, allow searching for any user ID. Entering the member ID 42 would hence list all programming groups where either internal user 42 or external user 42 is a member of. In that case, the variable should be `programming_group_memberships_user_id_eq`.
- Alternatively, provide two distinct search fields, one for internal and one for external users. This allows admins to search specifically. Then, we need two search fields, one named `programming_group_memberships_user_of_ExternalUser_type_id_eq` and the other `programming_group_memberships_user_of_InternalUser_type_id_eq`. This is called a polymorphic search and further outlined [in the docs](https://activerecord-hackery.github.io/ransack/going-further/polymorphic-search/). In this second case, we also need to allow searching for the user in the `model/programming_group_membership.rb` with
  ```rb
  def self.ransackable_associations(_auth_object = nil)
    %w[user]
  end
  ```
  and further add the `id` to `ransackable_attributes` in `model/user.rb`.

With any of these two methods, I was able to search successfully for corresponding users.

MrSerth:
> Based on ExternalUser exercise statistics (might be another ticket):
> 
> - programming_groups/:id/statistics

Feel free to create a ticket for this case, so that we can resolve my previous comment.

app/models/programming_group.rb
```diff
@@ -83,6 +83,10 @@ def users=(users)
     end
   end
 
+  def self.ransackable_associations(_auth_object = nil)
+    %w[exercise]
```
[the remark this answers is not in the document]
MrSerth: I am still leaving this thread open as a reminder to verify later (once the filters are finished). In the current state, we wouldn't need the `programming_group_memberships`. However, when adding the filtering based on user ID, it is required.

MrSerth:
I've pushed two minor commits, which fix issues with updating programming groups. One of them is dedicated to support groups with internal users (and probably was changed by mistake; we had this working already) and the other one is preventing a bug that would always recreate the membership in case of any (successful) update attempt. Since the second commit d11f487c is also changing the way new groups are created, I would kindly ask you to verify this change thoroughly.

app/views/programming_groups/index.html.slim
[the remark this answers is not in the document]
MrSerth: Awesome, thank you! Works perfectly fine 💪

MrSerth:
Nice done! I like the changes very much and also your effort of making the programming groups pages clickable within the breadcrumbs. I've taken your existing code and just refined it a little, to make it more generic (see 0e4b0ec4). This allows us to reuse it in more places, even when the parent resource is no longer an exercise. I've then used that mechanism for RfCs, even though they don't have a completely nested resource yet (767faf40).

Further, while performing these changes, there was one small convenience change I just added as well, which is the way we set the title for programming groups listed for a dedicated exercise (328d52eb).

Feel free to give these changes a final look as well 😉.

MrSerth (approved):
Awesome, well done! 💪

Issue #1914, opened 2023-09-18
WebSocket connection closed without (our custom) closing frame in case of a timeout


mpass99:
See [CODEOCEAN-TW](https://codeocean.sentry.io/issues/4459598724/)

<hr>

After running into a runner or execution timeout we should send the timeout message to CodeOcean. This behavior is tested by the unit tests `TestExitHasTimeoutErrorIfRunnerTimesOut` and `TestReturnsAfterTimeout`, and the e2e test `TestCommandReturnsAfterTimeout`.

<hr>

CodeOcean shows the wrong error `Alle Ausführungsumgebungen sind momentan in Benutzung. Probiere es später nochmal.`

mpass99:
By changing the debug level of CodeOcean to `debug`, I was able to verify that CodeOcean receives the timeout message.

![Image](https://github.com/openHPI/codeocean/assets/22845248/2bfcd547-b2cf-47b9-9694-6b1e8fd00e45)

I tried to debug the behavior of CodeOcean, but unfortunately I was not even able to get added debug statements in `connection.rb` to output :frowning_face:

MrSerth:
Awesome, thanks for having a further look there! I thought I had a look there, but obviously I didn't.

When investigating now, I identified that the JSON response sent by Poseidon is not recognized as valid and hence not accepted:

```rb
BACKEND_OUTPUT_SCHEMA.valid?({"type": "timeout"})
=> false

BACKEND_OUTPUT_SCHEMA.valid?({"type": "start"})
=> false
```

```rb
BACKEND_OUTPUT_SCHEMA.valid?({"type": "exit", data: 0})
=> true

BACKEND_OUTPUT_SCHEMA.valid?({"type": "stdout", data: "test"})
=> true

BACKEND_OUTPUT_SCHEMA.valid?({"type": "stderr", data: "test"})
=> true

BACKEND_OUTPUT_SCHEMA.valid?({"type": "error", data: "test"})
=> true
```

[...]

PR #1916, opened 2023-09-19
Fix Poseidon JSON schema

mpass99:
that included a wrong usage of the `required` attribute.

See #1914 and openHPI/poseidon#456.


MrSerth (approved):
Awesome, thanks! I just added a regression spec, please have another look.

mpass99:
Great opportunity to add a regression spec :+1:

PR #1925, opened 2023-09-22
Add custom redirect to LimeSurvey through LTI

MrSerth:
This PR adds the possibility to redirect learners to LimeSurvey (coming from LTI).

Further, this PR adds some general associations in the first commit (unrelated to the survey redirect, but required for the parameters added).

⚠️ I have tested the implementation only roughly, so please do that as part of the review.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -17,6 +17,8 @@ class SessionsController < ApplicationController
   end
 
   def create_through_lti
+    return redirect_to_survey if params[:custom_survey_id]
+
     session.delete(:pg_id) # Remove any previous pg_id from the session
     store_lti_session_data(params)
     store_nonce(params[:oauth_nonce])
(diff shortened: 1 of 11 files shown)


app/controllers/sessions_controller.rb
```diff
@@ -66,4 +68,34 @@ def destroy
...
+      .delete_if {|key, _| key.to_s.match(/.*(tracking|referrer|user_id).*/) }
+      .merge(
+        r: 'survey/index', # required LimeSurvey path passed as query param
+        sid: params[:custom_survey_id], # required LimeSurvey survey ID
+        newtest: 'Y', # force a new LimeSurvey session
+        xi_platform: 'openhpi' # pass a platform identifier
+      ).tap do |qp|
+      if current_user
+        # add a user pseudo ID if applicable
+        qp[:xi_pseudo_id] = Digest::SHA256.hexdigest(current_user.external_id)
```
selinareinhard, line 89: It would be nice to also have the study_group_id

MrSerth: I've just amended my previous commit and added the `study_group_id` as `co_study_group_id`.

kiragrammel:
Thank you for the changes! I tested it and everything works. 👍🏻

PR #1955, opened 2023-10-13
Undo Pair Programming Study

kiragrammel:
* Remove PairProgrammingExerciseFeedback
* Remove PairProgramming23Study and instead use LTI parameter
* Remove info about study & link to feedback page
* Replace "Du" with "Sie"
* Set bonus points for survey via LTI parameter

TODO:
- [ ] Remove `config/pair_programming23_study.csv` file from server


app/models/pair_programming_exercise_feedback.rb
```diff
@@ -12,6 +12,32 @@ class PairProgrammingExerciseFeedback < ApplicationRecord
   scope :intermediate, -> { where.not(normalized_score: 1.00) }
   scope :final, -> { where(normalized_score: 1.00) }
 
+  enum comment: {
+    difficulty_easy: 0,
```
kiragrammel, line 16: Should the link to the answer options in the locals remain or should the texts also be dragged into the model?

app/models/pair_programming_exercise_feedback.rb
```diff
@@ -12,6 +12,32 @@ class PairProgrammingExerciseFeedback < ApplicationRecord
...
+    difficulty_some_what_easy: 1,
+    difficulty_ok: 2,
+    difficulty_some_what_difficult: 3,
+    difficult_too_difficult: 4,
+  }, _prefix: true
+
+  enum time: {
+    estimated_time_less_5: 0,
```
kiragrammel, line 24: Rubocop  agrees with the following names: 
`estimated_time_less5`, 
    `estimated_time5to10`,
    `estimated_time10to20`,
    `estimated_time20to30`,
    `estimated_time_more30`
    
Should I rename it that way and then adjust the names in the locales as well?

config/locales/de.yml
```diff
@@ -994,23 +994,18 @@ de:
     difficulty_ok: "Die Aufgabenschwierigkeit war genau richtig."
     difficulty_some_what_difficult: "Die Aufgabe war etwas zu schwierig."
     difficult_too_difficult: "Die Aufgabe war viel zu schwierig."
-    difficulty: "Schwierigkeit der Aufgabe:"
```
kiragrammel, line 997: Should I remove the answer options as well?
And should I remove the lines (232 - 234)?
```
pair_programming_exercise_feedback:
        one: Feedback
        other: Feedbacks
```

app/controllers/exercises_controller.rb
```diff
@@ -315,9 +315,11 @@ def implement # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedC
       # we are just acting on behalf of a single user who has already worked on this exercise as part of a programming group **in the context of the current study group**
       session[:pg_id] = pg.id
       @current_contributor = pg
-    elsif session[:pg_id].blank? && PairProgramming23Study.participate_in_pp?(current_user, @exercise) && PairProgramming23Study::FORCED_EXERCISE_IDS.include?(@exercise.id)
+    # TODO: When the user is changing the url the session parameter is not set for the exercise
```
kiragrammel, line 318: I think it is only possible to get the value `pair_programming` for an exercise if we store it in the db or have it in a separate file. I think it is not wanted that we store it in the exercise db because the value can be different for different courses.

Maybe we shouldn't allow normal users to change the page via the URL and therefore just stay on the page when `session[:pg_id].blank?` is true. Then of course we can't save events for the different cases. What do you think?

app/models/pair_programming_exercise_feedback.rb
MrSerth: Mh, I see. In this case, I actually don't agree with Rubocop, because I would rate the consistency more important. Let's disable rubocop with a magic comment:

```rb
# rubocop:disable Naming/VariableNumber
```

Later on, re-enable rubocop again:

```rb
# rubocop:enable Naming/VariableNumber
```

app/models/pair_programming_exercise_feedback.rb
MrSerth: I would propose to remove them from the locales as well, so that we _just_ have the model and the database table.

When doing so, I would recommend adding a comment to the enum if the label chosen is not clear enough. Then, a future researcher can still gather meaningful insights from the data.

app/models/pair_programming_exercise_feedback.rb
```diff
@@ -12,6 +12,32 @@ class PairProgrammingExerciseFeedback < ApplicationRecord
   scope :intermediate, -> { where.not(normalized_score: 1.00) }
   scope :final, -> { where(normalized_score: 1.00) }
 
+  enum comment: {
```
MrSerth, line 15: If you match the column name here (please do so), Rails will do some magic 🙌 .

Then, you can also remove the prefix `difficulty` from the answer options.

app/models/pair_programming_exercise_feedback.rb
MrSerth, line 16: Let's go with `too_easy`, since this would match the translation (and the `too_difficult`) better.

app/models/pair_programming_exercise_feedback.rb
```diff
@@ -12,6 +12,32 @@ class PairProgrammingExerciseFeedback < ApplicationRecord
...
+    difficulty_easy: 0,
+    difficulty_some_what_easy: 1,
+    difficulty_ok: 2,
+    difficulty_some_what_difficult: 3,
+    difficult_too_difficult: 4,
+  }, _prefix: true
+
+  enum time: {
```
MrSerth, line 23: Same as above, let's use `user_estimated_worktime`

app/models/pair_programming_exercise_feedback.rb
MrSerth, line 24: Then, for those names, the prefix `estimated_time` can be removed. I am still thinking whether we should add the unit `minutes` (or a short form `min`) to the name, just to make the translations obsolete.

app/models/pair_programming_exercise_feedback.rb
```diff
@@ -12,6 +12,32 @@ class PairProgrammingExerciseFeedback < ApplicationRecord
...
+    estimated_time_less_5: 0,
+    estimated_time_5_to_10: 1,
+    estimated_time_10_to_20: 2,
+    estimated_time_20_to_30: 3,
+    estimated_time_more_30: 4,
+  }, _prefix: true
+
+  enum reasons: {
```
MrSerth, line 31: With Rails magic:

```suggestion
  enum reason_work_alone: {
```

And then, we can remove the `reason` prefix from the entries below.

app/models/pair_programming_exercise_feedback.rb
```diff
@@ -12,6 +12,32 @@ class PairProgrammingExerciseFeedback < ApplicationRecord
...
+    reason_no_partner: 0,
+    reason_to_difficult_to_find_partner: 1,
```
MrSerth, line 33:
```suggestion
    too_difficult_to_find_partner: 1,
```

app/models/pair_programming_exercise_feedback.rb
```diff
@@ -12,6 +12,32 @@ class PairProgrammingExerciseFeedback < ApplicationRecord
...
+    reason_faster_alone: 2,
+    reason_not_working_with_strangers: 3,
+    reason_want_to_work_alone: 4,
```
MrSerth, line 36:
```suggestion
    prefer_to_work_alone: 4,
```

config/locales/de.yml
MrSerth: I would say so, yes (for both, the answer options and the model human name).

app/controllers/exercises_controller.rb
MrSerth: > I think it is not wanted that we store it in the exercise db because the value can be different for different courses.

Indeed, that's why we pass the option through LTI.

> Maybe we shouldn't allow normal users to change the page via the URL and therefore just stay on the page when `session[:pg_id].blank?` is true. 

Sure, this is something we could consider (and might have positive consequences). However, for now, I would not suggest extending this beyond the original scope of this PR. Therefore, I would go with a similar solution to what we had so far.

<hr>

That having said, I think the solution currently in the code is okay. Sure, there is still the issue that the session parameter could still be circumvented (e.g., by starting an exercise without mandatory Pair Programming and then switching the URL to a mandatory Pair Programming exercise). However, a user cannot receive any points through this manipulation (but just see the exercise). Hence, I would accept this for now.

In that regard, I am not sure whether I get the code comments correctly. Did I answer that with my comment already or is there another question I missed (and didn't get)?

app/controllers/concerns/lti.rb
```diff
@@ -226,6 +226,7 @@ def store_lti_session_data(parameters)
     @lti_parameters.save!
 
     session[:external_user_id] = current_user.id
+    session[:pair_programming] = params[:custom_pair_programming] || false
```
MrSerth, line 229: Where are we removing this variable again? At some point, we should probably do so.

MrSerth (commented):
Nice, thank you. The removal looks good so far, I just made a few minor comments.

app/controllers/concerns/lti.rb
MrSerth, line 229: Since this concern should not have direct access to the parameters, use the variable passed as a parameter instead:

```suggestion
    session[:pair_programming] = parameters[:custom_pair_programming] || false
```

MrSerth: Further, we can probably expect that we store this information in the session. Within the respective spec, add:

```rb
expect(controller.session).to receive(:[]=).with(:pair_programming, anything)
```

MrSerth (commented):
Here's a hint for the failing tests.

app/controllers/exercises_controller.rb
kiragrammel: You answered all questions. I overlooked that the user then cannot get points. Then it should be ok like this for now.

app/models/pair_programming_exercise_feedback.rb
MrSerth: With the current version, the naming for less than 5 minutes has an additional underscore before the unit, which all other keys are missing. Shouldn't we use the same naming scheme here?

MrSerth: As I see, adding a unit to those values also fixed the Rubocop annotation without disabling the `Naming/VariableNumber` rule 💪

PR #1982, opened 2023-10-29
Exclusively lock Runners during code executions

MrSerth:
Previously, the same runner could be used multiple times with different submissions simultaneously. This, however, yielded errors, for example when one submission time oud (causing the running to be deleted) while another submission was still executed.

Admin actions, such as the shell, can be still executed regardless of any other code execution.

Fixes CODEOCEAN-HG
Fixes openHPI/poseidon#423

### DIFF
--- a/app/controllers/live_streams_controller.rb
+++ b/app/controllers/live_streams_controller.rb
@@ -26,15 +26,15 @@ class LiveStreamsController < ApplicationController
     runner = Runner.for(current_user, @execution_environment)
     fallback_location = shell_execution_environment_path(@execution_environment)
     privileged = params[:sudo] || @execution_environment.privileged_execution?
-    send_runner_file(runner, desired_file, fallback_location, privileged:)
+    send_runner_file(runner, desired_file, fallback_location, privileged:, exclusive: false)
   end
 
   private
 
-  def send_runner_file(runner, desired_file, redirect_fallback = root_path, privileged: false)
+  def send_runner_file(runner, desired_file, redirect_fallback = root_path, privileged: false, exclusive: true)
     filename = File.basename(desired_file)
     send_stream(filename:, type: 'application/octet-stream', disposition: 'attachment') do |stream|
-      runner.download_file(desired_file, privileged_execution: privileged) do |chunk, overall_size, _content_type|
+      runner.download_file(desired_file, privileged_execution: privileged, exclusive:) do |chunk, overall_size, _content_type|
         unless response.committed?
           # Disable Rack::ETag, which would otherwise cause the response to be cached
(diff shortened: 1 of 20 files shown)


config/locales/de.yml
```diff
@@ -398,6 +398,7 @@ de:
       network: 'Während Ihr Code läuft, ist Port %{port} unter folgender Adresse erreichbar: <a href="%{address}" target="_blank" rel="noopener">%{address}</a>.'
       render: Anzeigen
       run: Ausführen
+      runner_in_use: Sie führen momentan Code aus. Bitte stoppen Sie die vorherige Ausführung oder warten Sie einen Moment, bevor Sie fortfahren.
```
mpass99, line 401: It seems that I cannot stop a score run (even though it is running longer than 30s). Is this expected?

mpass99 (commented):
Nice, It's a great control mechanism!

app/controllers/live_streams_controller.rb
```diff
@@ -26,15 +26,15 @@ def download_arbitrary_file
...
   end
 
   private
 
-  def send_runner_file(runner, desired_file, redirect_fallback = root_path, privileged: false)
+  def send_runner_file(runner, desired_file, redirect_fallback = root_path, privileged: false, exclusive: true)
     filename = File.basename(desired_file)
     send_stream(filename:, type: 'application/octet-stream', disposition: 'attachment') do |stream|
-      runner.download_file(desired_file, privileged_execution: privileged) do |chunk, overall_size, _content_type|
+      runner.download_file(desired_file, privileged_execution: privileged, exclusive:) do |chunk, overall_size, _content_type|
```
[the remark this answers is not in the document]
MrSerth: Indeed, it is meant as `exclusive: exclusive`. This is a new feature introduced in Ruby 3.1 as described [here](https://rubyreferences.github.io/rubychanges//3.1.html#values-in-hash-literals-and-keyword-arguments-can-be-omitted).

config/locales/de.yml
MrSerth: Currently, score runs cannot be stopped at all, that's correct. As of now, we are lacking a corresponding control mechanism, which clearly is a limitation. Since an exercise might contain multiple test files, the time to wait might be multiples of the permitted execution timeout: Consider an exercise with 2 test files. If each of the test files times out (and 30 seconds is set), you might need to wait a maximum of 2 (test files) x 30 seconds (timeout) = 1 minute. 

This behaviour is unchanged, but of course gets a little more annoying with blocking runners and no option to cancel the score run at all.

mpass99 (approved):

PR #2041, opened 2023-11-30
Fix Sentry span for outgoing WebSocket connections

MrSerth:
Recently, we noticed that Poseidon's WebSocket spans shown in Sentry are not correctly aligned with the outgoing WebSocket request in CodeOcean. This PR aims to fix that problem.

[Here's a sample trace with the code of this branch](https://codeocean.sentry.io/performance/codeocean:087ae8ca52084d06a5e25021bf7cc144/). 

Related to https://github.com/openHPI/poseidon/issues/438


mpass99 (approved):
Nice finding! Looks good 👍

PR #2066, opened 2023-12-22
Change syntax for routes with filename

MrSerth:
Previously, the filename was URL-encoded, thus each / was replaced with %2F. This caused issues with some Apache2 configuration, smartly mingling with the URL to either encode it a second time (resulting in %252F) or decoding it (generating a real /). However, for authenticated file downloads with the JWT, we hardly require a byte-by-byte matching. With these changes, the URL parameter is no longer URL-encoded, so that Apache2 won't break our implementation any longer.

Further, we use this opportunity to get rid of the unnecessary .json extension for those filename routes, simplifying the routes generated and doing some further cleanup.

<hr>

Finally, this PR fixes some other open bugs corresponding to file handling (either in the web server or the runner management), and cleans up some Frontend issues in dedicated commits.


MrSerth:
We verified these changes with an Apache2 server, confirming the efficacy of the changes. Hence, we are now ready to merge this PR.

PR #2139, opened 2024-02-13
Extract Contributor concern as abstract class

MrSerth:
During documentation of the pair programming feature, we noticed that the Contributor should be an abstract class, which is parent for the User and the ProgrammingGroup. With this commit, we perform these changes.

### DIFF
--- a/app/models/concerns/contributor.rb
+++ b/app/models/concerns/contributor.rb
@@ -1,9 +1,0 @@
-# frozen_string_literal: true
-
-module Contributor
-  extend ActiveSupport::Concern
-
-  included do
-    has_many :submissions, as: :contributor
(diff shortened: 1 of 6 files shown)


Dome-GER (approved):
Pretty clean changes, looks good based on the code. I did not cover this from the domain perspective.

MrSerth:
Alright, thanks for having another look. I think, it should be good to go, so let's give it a try 😁

MrSerth:
> No spec adjustments?

No, there are no specs adjustments (yet). The application behaviour is pretty unchanged; and we didn't want to just the inheritance (since this is not handled by our code). If I come up with another useful spec, I'll add one.

PR #2144, opened 2024-02-14
Use Webpack to deliver newest ACE editor

MrSerth:
With this PR, we change the delivery method for the ACE editor from manually copied files to using yarn and webpack. As a side-change, we also modify how the mode is selected through JavaScript instead of Ruby. 

I've tested these changes on CodeOcean Staging already, they work as intended.

<hr>

This PR aligns with openHPI/codeharbor#1240.


kiragrammel (approved):
Apologies for the late reply.
The changes look good from my point of view. 
Great result! 😊👍🏻

MrSerth:
Awesome, thanks for taking the time to validate the changes again and approving them - I really appreciate that! I would love to see this PR merged soon, allowing me to finalise the work here 😊.

Dome-GER (approved):
This is a huge step forward! Thanks.

PR #2145, opened 2024-02-14
Resolve current deprecation warnings

MrSerth:
This PR resolves two open deprecation warnings:

- `DEPRECATION WARNING: Calling `ActiveRecord::Base.clear_active_connections! is deprecated. Please call the method directly on the connection handler; for example: `ActiveRecord::Base.connection_handler.clear_active_connections!`. (called from clear_active_connections! at /home/runner/work/codeocean/codeocean/vendor/bundle/ruby/3.3.0/gems/activerecord-7.1.3/lib/active_record/connection_handling.rb:320)` is resolved in the first commit f50ff1f7489d25ed5d1ab4825c21e1e42661b4be.
- `Rails 7.1 has deprecated the singular fixture_path in favour of an array.You should migrate to plural` is resolved in the second commit 89fd9d193e059476ba333a6f9d9427128578ffa4.

### DIFF
--- a/config/initializers/monkey_patches.rb
+++ b/config/initializers/monkey_patches.rb
@@ -28,3 +28,25 @@ module Sorcery
     end
   end
 end
+
+# Tubesock uses a deprecated method to clear active connections with ActiveRecord.
+# Hence, we update the method call to the new method name (including `connection_handler`).
+module Tubesock::Hijack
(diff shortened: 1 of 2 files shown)


Dome-GER (approved):
Very good solution with the separate initializer.

PR #2190, opened 2024-03-12
fix kramdown using curved quotes 

kkoehn:
Kramdown uses curved quotation marks by default, when rendering markdown to html

This changes them to be straight quotes, so they can be easily copied in code or the like.


MrSerth (approved):
Thanks, looks good!

PR #2248, opened 2024-04-14
Introduce i18n-tasks and lint locales

MrSerth:
This PR adds an [i18n-tasks](https://github.com/glebm/i18n-tasks) configuration similar to CodeHarbor.
Further, it moves all locales to their new location and ensures we follow Rails conventions where possible.

_Hint: Reviewing commit-by-commit is highly recommended._


app/controllers/code_ocean/files_controller.rb
```diff
@@ -66,7 +66,7 @@ def create_and_respond(options = {})
         else
           filename = "#{@object.path || ''}/#{@object.name || ''}#{@object.file_type.try(:file_extension) || ''}"
           format.html do
-            flash[:danger] = t('files.error.filename', name: filename)
+            flash[:danger] = t('code_ocean/files.error.filename', name: filename)
```
Dome-GER, line 69: The prefix is interesting. I'm not really used to this, using `/`.

app/helpers/statistics_helper.rb
```diff
@@ -30,19 +30,19 @@ def contributor_statistics
     [
       {
         key: 'internal_users',
-          name: t('activerecord.models.internal_user.other'),
+          name: InternalUser.model_name.human(count: :many),
```
Dome-GER, line 33: Is it recommended doing this? For me, it's less understandable tbh.

Dome-GER (approved):
Very nice, clean commits. Especially the last one with using the helpers feels weird. I seem to prefer explicit locale paths.

app/controllers/code_ocean/files_controller.rb
MrSerth: Yes, I seldomly use it, too. However, since the class is called `CodeOcean::File` (and not just `File` due to the conflict with the internal class representing access to objects on the filesystem), we need to use it everywhere... This is actually the prefix Rails expect if using the `CodeOcean::File.model_name.human` method. 🤷

app/helpers/statistics_helper.rb
MrSerth: Are you referring to the `model_name.human` or the `count: :many`?

Regarding the `count: :many`: I would say so. Previously, we sometimes had `count: 2`, but in some languages with more complex pluralization forms, this wouldn't be appropriate. However, I would also be fine to go with `count: :other`, if you think that's better. The explanation I am aware of is:

- `:zero`: Used when the count is zero.
- `:one`: Used when the count is one.
- `:two`: Used in languages where a specific form exists for the count of two.
- `:few`: Used in languages with a specific form for a small number of items, typically around 3-4.
- `:many`: Used for larger quantities, typically used for numbers that are large but not necessarily unlimited.
- `:other`: Used as a general plural form when none of the other rules apply.

Regarding the `model_name.human` (vs `t('activerecord.models.')`), I think we could discuss that. While the form `model_name.human` seems to be more complex and longer, I actually prefer it. My main reason would be that it automatically enforces the correct alignment in the locales, requiring a pre-defined convention to be used. Previously, this wasn't always ensured, as the example with `code_ocean/files` showed (that worked with the manual translation lookup, but failed with the `model_name.human` approach).

<details><summary>Details</summary>
<p>

I also tasked ChatGPT 4o to provide some details:

> Where can I find some information on the difference between `.model_name.human` in Rails and explicit `t('activerecord.models.')` in Rails?

<hr>

In Ruby on Rails, both `.model_name.human` and `t('activerecord.models.')` are used to retrieve the human-readable names for ActiveRecord models, but they differ in their usage and context.

### `.model_name.human`

`.model_name.human` is a method provided by ActiveRecord. It retrieves the human-readable name for a model based on the I18n translations. This method leverages Rails' built-in conventions and is convenient for quickly accessing the model's name without directly referencing translation keys.

- **Usage Example:**
  ```ruby
  User.model_name.human
  ```

- **Behavior:**
  This method will look up the translation key `activerecord.models.user` in the locale files. If a translation for the specific model is not found, it will fallback to a titleized version of the model name (e.g., "User" for `User` model).

- **Advantage:**
  It simplifies the process of getting the human-readable model name by encapsulating the logic within the method, reducing the need to directly reference translation keys.

### `t('activerecord.models.')`

`I18n.t('activerecord.models.')` (or simply `t('activerecord.models.')` in views) is a more explicit way of fetching translations. This method directly references the translation keys in your locale files.

- **Usage Example:**
  ```ruby
  t('activerecord.models.user')
  ```

- **Behavior:**
  This method explicitly looks up the given key in the locale files. It requires you to know the exact key and use it directly, which can be beneficial for clarity and control.

- **Advantage:**
  Provides explicit control over which translation key is being accessed. This can be useful in complex applications where you need precise control over translations or when working with nested or custom keys.

### Key Differences

1. **Convenience vs. Explicit Control:**
   - `.model_name.human` is more convenient and abstracts away the details of translation key lookup.
   - `t('activerecord.models.')` provides explicit control, allowing for more direct and precise access to translation keys.

2. **Fallback Mechanism:**
   - `.model_name.human` has a built-in fallback to the titleized model name if the translation is missing.
   - `t('activerecord.models.')` will raise an error if the translation is missing unless a fallback is explicitly provided using the `:default` option.

3. **Usage Context:**
   - `.model_name.human` is typically used within model contexts or when you need a quick and easy way to get the model name.
   - `t('activerecord.models.')` is often used in views and controllers where more complex or nested translation keys might be involved.

### Example Locale File

Here’s how you might define translations in your `config/locales/en.yml`:

```yaml
en:
  activerecord:
    models:
      user: "User"
    attributes:
      user:
        name: "Name"
```

### Example Usage in Rails

- Using `.model_name.human`:
  ```ruby
  class User < ApplicationRecord
    def display_name
      "#{self.class.model_name.human}: #{name}"
    end
  end
  ```

- Using `t('activerecord.models.user')` in a view:
  ```erb
  <%= t('activerecord.models.user') %>
  ```

In summary, use `.model_name.human` for convenience and less boilerplate, while `t('activerecord.models.')` gives you more explicit control and flexibility in accessing translation keys.

</p>
</details>

MrSerth: I updated the usage from `:many` to `:other`.

PR #2249, opened 2024-04-14
Add Slim Lint

MrSerth:
This PR adds slim-lint to the code base, and CI. Further, it fixes all current code issues.


Dome-GER:
Very nice! I was thinking about introducing it for Xikolo as well, but I'm scared because of the high number of violations..

MrSerth:
> Very nice! I was thinking about introducing it for Xikolo as well, but I'm scared because of the high number of violations..

While performing these changes, I actually noticed that we already have a `.slim-lint.yml` config in web since two years. So, I would say it's time to enforce it 🫣.

PR #2252, opened 2024-04-14
Connection::Buffer: Ignore linter indicator accidentally wrapped in quotes

MrSerth:
The PyLint output is marking the erroneous location visually by using ^ in the next line. This indicator is always prefixed by multiple spaces and the line does not contain any further text. Now, it might happen that the indicator is between two source lines that use the same quote type. Since we detect quotes through our regex, this occurrence would qualify, effectively enquoting the linter indicator ^. As a consequence of our quote handling, we normally prevented quoted text from being split. However, in the described scenario, this is not desired, since we are not dealing with a regular quote (but an accidental mismatch).

Therefore, with this commit, we disable our quote handling for those multi-line quotes where one line fully represents a typical PyLint indicator line. This should restore the desired line-matching behavior.

Fixes [CODEOCEAN-12N](https://codeocean.sentry.io/issues/5161130664/)


MrSerth:
Here is a review aid:

**Previous Line Splitting**
Bildschirmfoto 2024-04-14 um 23 27 53

**Updated Line Splitting**
Bildschirmfoto 2024-04-14 um 23 27 22

Dome-GER (approved):

PR #2258, opened 2024-04-17
RfCs: Allow filtering for all distinct states

MrSerth:
Previously, the select for RfCs only allowed filtering for two states:

Bildschirmfoto 2024-04-17 um 21 00 29

In this solution, "unsolved" RfCs contained ongoing and soft-solved RfCs. Now, we allow filtering for three states:

Bildschirmfoto 2024-04-17 um 21 00 05

<hr>

Further, we update the colors:

| Before | After |
|--------|--------|
| Bildschirmfoto 2024-04-17 um 21 01 36 | Bildschirmfoto 2024-04-17 um 21 02 22 |


Dome-GER (approved):

app/views/request_for_comments/show.html.slim
```diff
@@ -1,7 +1,9 @@
 .list-group
   h4#exercise_caption.list-group-item-heading data-exercise-id=@request_for_comment.exercise.id data-rfc-id=@request_for_comment.id
     - if @request_for_comment.solved?
-      span.fa-solid.fa-check.me-2 aria-hidden='true'
+      span.fa-solid.fa-check.text-success.me-2 aria-hidden='true'
+    - elsif @request_for_comment.full_score_reached
```
Dome-GER, line 5:
```suggestion
    - elsif @request_for_comment.full_score_reached?
```

Sounds like a boolean condition, so we can indicate it accordingly. Didn't look it up.

Dome-GER (approved):
Minor improvements left.

config/locales/de.yml
```diff
@@ -857,6 +857,7 @@ de:
     mark_as_solved: "Diese Frage als beantwortet markieren"
     show_all: "Alle Anfragen anzeigen"
     show_solved: "Nur gelöste Anfragen anzeigen"
+    show_soft_solved: "Nur potentiell gelöste Anfragen anzeigen"
```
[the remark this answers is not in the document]
MrSerth: Ah, thanks. Didn't notice ;)

app/views/request_for_comments/show.html.slim
MrSerth: Yes, that's a boolean condition. I verified the change locally, and as expected and usual in these cases, the `?` can be appended.

PR #2288, opened 2024-05-08
Change order of Sprocket assets and introduce dependency

MrSerth:
This PR aims to fix another JavaScript error shown in Sentry: [CODEOCEAN-FRONTEND-7W](https://codeocean.sentry.io/issues/5276691444/).

The error indicates that `CodeOceanEditorEvaluation is not defined` in the following location:

https://github.com/openHPI/codeocean/blob/7d5883acc4f8a7af8a105fabdbd99410b7673e2a/app/assets/javascripts/editor.js#L1-L14

I was surprised by this error and have never seen it before. Still, it is similar to [CODEOCEAN-FRONTEND-7D](https://codeocean.sentry.io/issues/5237549093/) fixed in #2270. Thus, I started to reflect on why that happens and what could be the issue. 

One idea I got is linked to the recent changes we made with JavaScript assets: We changed how the ACE editor and the Markdown editor were included, removing them from Sprockets (and its general `application.js` bundle), and moved them over to Webpack. Effectively, this reduced the file size of the Sprockets-generated JavaScript file, and potentially modified the order of elements. Potentially, this could result in a race condition, as previously identified with #2270 already. Now, however, I think that the way we generate the Sprockets-based  `application.js` bundle is the issue: By just using `require_tree`, we effectively included all our JavaScript files in an alphabetical order [[1]](https://github.com/rails/sprockets?tab=readme-ov-file#file-order-processing). This hasn't changed, and removing items should not affect the remaining alphabetical order. However, we also use Turbolinks with the `turbolinks:load` event.

The `turbolinks:load` event "fires once after the initial page load, and again after every Turbolinks visit" [[2]](https://github.com/turbolinks/turbolinks?tab=readme-ov-file#full-list-of-events). Further a contributor states that "`turbolinks:load` is always fired when Turbolinks loads a page via AJAX, but it does not wait for scripts to load" [[3]](https://github.com/turbolinks/turbolinks/issues/534#issuecomment-613064059). Investigating on our bundled `application.js`, I identified that the problematic `CodeOceanEditorEvaluation` (as well as other `CodeOceanEditor` elements) were only declared after being used in one of the `turbolinks:load` handlers. A good example for that is `CodeOceanEditorAJAX`, as shown in the table.

Thus, my conclusion is that the alphabetical order chosen by Sprockets is not ideal for our case, and maybe the smaller bundle size makes it more likely that we are now seeing the respective issues. As a consequence, I manually checked the dependencies and extracted a few components that I think should be loaded first. Once these are included in the bundle, I still use the `require_tree`, since Sprockets will automatically de-duplicate entries [[4]](https://github.com/rails/sprockets?tab=readme-ov-file#require) to ensure each asset is only included once in the bundle.

| Before | After |
|--------|--------|
| Bildschirmfoto 2024-05-08 um 23 43 29 | Bildschirmfoto 2024-05-08 um 23 44 08 | 
| `CodeOceanEditorAJAX ` is first used before declared | `CodeOceanEditorAJAX ` is first declared, later used (search: `3 of 3`) |

Please take a moment to think through the solution chosen and whether you'd assume it fixes the underlying issue. Further, reviewing commit-by-commit is recommended.

### DIFF
--- a/lib/code_ocean/config.rb
+++ b/lib/code_ocean/config.rb
@@ -2,12 +2,16 @@
 
 module CodeOcean
   class Config
-    def initialize(filename)
-      @filename = filename
+    attr_reader :path, :read
+
+    def initialize(filename, options = {})
+      @path = Rails.root.join('config', "#{filename}.yml#{options[:erb] ? '.erb' : ''}")
+      @read = parse(options)
     end
 
-    def read(options = {})
-      path = Rails.root.join('config', "#{@filename}.yml#{options[:erb] ? '.erb' : ''}")
+    private
+
+    def parse(options)
       if ::File.exist?(path)
         yaml_content = ::File.new(path, 'r').read || ''
         yaml_content = ERB.new(yaml_content).result if options[:erb]
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -19,4 +19,19 @@
 //= require color_mode_picker
 //
 // app/assets
+// --> Include some assets first, as they are used by other assets.
+// --> Hence, the order specified here is important.
+//
+// 1. Some common base functions and monkey patches
+//= require base
+// 2. Programming groups are required by "channels/synchronized_editor_channel.js"
+//= require programming_groups
+// 3. The turtle library is required by "editor/turtle.js"
+//= require turtle
+// 4. Some channels are required by "editor/editor.js.erb"
+//= require_tree ./channels
+// 5. Require the editor components, as needed by "./editor.js" and "./community_solution.js"
+//= require_tree ./editor
+//
+// All remaining assets are loaded in alphabetical order
 //= require_tree .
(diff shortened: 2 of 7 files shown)


lib/code_ocean/config.rb
```diff
@@ -2,12 +2,18 @@
 
 module CodeOcean
   class Config
-    def initialize(filename)
+    attr_reader :filename, :path, :read, :options
+
+    def initialize(filename, options = {})
       @filename = filename
+      @options = options
```
Dome-GER, line 9: If the options and filename are not used (yet), I tend to not expose them.

MrSerth: 👍, I've changed this and only export the `path` and content (via `read`) now.

MrSerth:
> Complex stuff. I agree it's not a beauty but might fix the issue. 🤞

Yes, let's see. I'll merge this PR and observe changes to Sentry.

> In the mid/long-term, i think limiting the JS and passing required data in via data attributes is a desired way.

I fully agree; that's really a long-term goal to refactor some of the JavaScript involved here.

MrSerth:
I've deployed the changes a few minutes ago: [Release 61513cc5bcc96171a99d73a34eeeafc200ec60f2](https://codeocean.sentry.io/releases/61513cc5bcc96171a99d73a34eeeafc200ec60f2/?project=5667283). Let's observe it...

PR #2312, opened 2024-05-24
Explicitly state permissions for CommunitySolutionPolicy

MrSerth:
Viewing the index page for community solutions was broken, since the breadcrumbs tried to access the `show?` policy.


Dome-GER (approved):
My inner monk is crying because of the order of the actions. 😇

MrSerth:
> My inner monk is crying because of the order of the actions. 😇

Thanks for pointing that out, I restored the default order :+1:

PR #2321, opened 2024-05-25
Prevent duplicate error handling for WebSocket events

MrSerth:
With the current implementation, some errors were captured twice (or only a single time, even if not desired at all).

The special WebSocket errors are not nicely captured here:

Bildschirmfoto 2024-05-25 um 20 15 00

Further, they lacked a complete stack trace:

Bildschirmfoto 2024-05-25 um 20 16 21

Therefore, we delegate this to a dedicated method `showWebsocketError` already capturing Sentry events:

https://github.com/openHPI/codeocean/blob/c5518bb592aa47650db29ed2cc5b8e1862853dab/app/assets/javascripts/editor/editor.js.erb#L895-L905

Regularly, we disable this error tracing, i.e., when closing the connection.

https://github.com/openHPI/codeocean/blob/c5518bb592aa47650db29ed2cc5b8e1862853dab/app/assets/javascripts/editor/evaluation.js#L144

Still, with the current implementation, a (misleading) error would still be captured a single time in Sentry.

### DIFF
--- a/app/assets/javascripts/editor/editor.js.erb
+++ b/app/assets/javascripts/editor/editor.js.erb
@@ -215,6 +215,9 @@ var CodeOceanEditor = {
                     try {
                         return await callback();
                     } catch (error) {
+                        // WebSocket errors are handled in `showWebsocketError` already.
+                        if (error.target instanceof WebSocket) return;
+
                         console.error(JSON.stringify(error));
                         Sentry.captureException(JSON.stringify(error), {mechanism: {handled: false}});
                     }


Dome-GER (approved):

jgraichen:
I was going to ask if you thought about moving error reporting to Sentry to the catch handler, and keep `showWebsocketError()` at only _showing_ the error to the user?

MrSerth:
> I was going to ask if you thought about moving error reporting to Sentry to the catch handler, and keep `showWebsocketError()` at only _showing_ the error to the user?

I considered that, yes. However, the event thrown by a WebSocket error does not contain any stack trace or other context (besides the `target`). Thus, an error would always be reported as originated from this method, which I disliked. Therefore, I decided to keep it in the other method, which will automatically annotate the function name in Sentry.

PR #2384, opened 2024-07-01
Fix missing/incorrect Model specifications in migrations

MrSerth:
This PR fixes one issue:
- It ensures we can run `db:migrate:reset` on an empty database by ensuring the model classes are specified correctly.

Further, the schema itself is fine and does not contain any major issues. There is one minor problem with the order of columns, that has not been solved yet. The current schema represents the production setup, but regenerating the schema for an empty database yields a slightly different order. For now, I did not touch this (since it worked previously without any problem).

_PS: While the migrations can now run successfully, I have not validated or touched them. Probably, some could have been designed better._

### DIFF
--- a/db/migrate/20140826073318_sorcery_core.rb
+++ b/db/migrate/20140826073318_sorcery_core.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 class SorceryCore < ActiveRecord::Migration[4.2]
+  class InternalUser < ApplicationRecord
+  end
+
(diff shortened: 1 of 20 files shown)


Dome-GER (approved):
You are doing quite some stuff in your migrations ;)

MrSerth:
Indeed, since we are sometimes using migrations to transform existing data (which could be done using a Background Job, too, for which we lack the necessary infrastructure so far).

PR #2393, opened 2024-07-08
Correctly set `hidden_feedback` for inherited files

MrSerth:
Without this fix, `hidden_feedback` was only effective when `read_only` was set, too.

/cc @brry: This fixes the erroneous behavior we just observed together.


brry:
Nice to see immediate bugfixes, as usual :)
Thanks!

Dome-GER (approved):

PR #2442, opened 2024-08-01
Fix memorization for false values

MrSerth:
Previously, a boolean value of `false` would not have been memorized correctly. Instead, the mechanism commonly used with `||=` would always reevalute the original, rendering the memorization useless. `true` values were not affected of this bug.

### DIFF
--- a/app/models/runner.rb
+++ b/app/models/runner.rb
@@ -19,13 +19,13 @@ class Runner < ApplicationRecord
   end
 
   def self.management_active?
-    @management_active ||= begin
-      runner_management = CodeOcean::Config.new(:code_ocean).read[:runner_management]
-      if runner_management
-        runner_management[:enabled]
-      else
-        false
-      end
+    return @management_active if defined? @management_active
+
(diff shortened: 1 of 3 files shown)


app/models/runner.rb
```diff
@@ -19,13 +19,13 @@ def self.strategy_class
...
 
   def self.management_active?
-    @management_active ||= begin
-      runner_management = CodeOcean::Config.new(:code_ocean).read[:runner_management]
-      if runner_management
-        runner_management[:enabled]
-      else
-        false
-      end
+    return @management_active unless @management_active.nil?
```
Dome-GER, line 22: How about `defined?(@management_active)`? 🤔 Could be a bit more clear..

Dome-GER: As discussed in person, I don't have a strong opinion here.

MrSerth: Thanks for your input. As discussed, I went with `defined?` to cover `nil` values as well and have the respective code part easier to read (with an `if` rather than an `unless`).

Dome-GER (approved):

PR #2447, opened 2024-08-03
Allow teachers most reading actions on CodeOcean

MrSerth:
With these changes, teachers gain more permissions on CodeOcean. Besides reading many objects, they can interact with execution environments more by now. For index pages, teachers only see those objects they are allowed to show. Furthermore, this PR:

- resolves a bug regarding the number of users for a study group when signed in as a teacher
- enables further filter for teachers (especially to find external and internal users better)
- enables teachers (and admin) to edit comments to RfCs, i.e. in case of explicit comments

Reviewing commit-by-commit is recommended

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -19,7 +19,7 @@ class CommentsController < ApplicationController
         comment.username = comment.user.displayname
         comment.date = comment.created_at.strftime('%d.%m.%Y %k:%M')
         comment.updated = (comment.created_at != comment.updated_at)
-        comment.editable = comment.user == current_user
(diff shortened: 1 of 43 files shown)


Dome-GER:
This still looks very complex, especially since I'm not familiar with your policy model. 😇

MrSerth:
> This still looks very complex, especially since I'm not familiar with your policy model. 😇

I agree, it's not simple. Having some knowledge with Pundit is beneficial (for the policies and the scopes) to review. Further, some aspects could be simplified -- I second that. For now, I would proceed with the current PR (it is an improvement compared to the status quo), and we might continue refactoring the policies later 😇.

PR #2455, opened 2024-08-09
Fix working time statistics when viewed without autosaves

MrSerth:
Previously, hiding autosaves also mixed up with the calculation of working times. This was caused by the missing submissions not being reflected in the deltas. The solution implemented here is not nice, but fulfills its requirements. Refactoring of the whole view is recommended.

| With Autosaves | Before | After |
|--------|--------|--------|
| Bildschirmfoto 2024-08-09 um 13 26 14 | Bildschirmfoto 2024-08-09 um 13 26 27 | Bildschirmfoto 2024-08-09 um 13 26 36 |


Dome-GER (approved):
Ouf. A hack for a not so nice piece of code.

PR #2478, opened 2024-08-26
Upgrade to Rails 7.2, switch to system specs, and add PWA and devcontainers support

MrSerth:
With this PR, we are upgrading CodeOcean to Rails 7.2 and making use of all the great features the new version provides. Therefore, this PR looks rather big at a first glance, but the commits are intended to provide a meaningful grouping of the changes. All commits are required to complete the upgrade and the activation of current best-practices for Rails application.

From a high-level point of view, the following changes are relevant:
- CodeOcean is running Rails 7.2 (instead of Rails 7.1)
- Remove deprecated `secrets.yml`, require setting environment variable `SECRET_KEY_BASE` for production
- Add brakeman as security scanner (and ignore some false positives)
- Rspec feature specs are migrated to system specs, which allows the usage of advanced features (automatic screenshots in case of failures, easier management of Capybara, no need to use the DatabaseCleaner gem any longer)
- Applying best-practices for Rails and Rspec, reordering some files
- Adding a webmanifest for CodeOcean, which allows to install CodeOcean as a PWA
- Chrome is used as a new default browser for system specs (rather than Firefox)
- Since the new system specs with Chrome are faster than the previous ones with Firefox, we need to wait explicitly for pending Websocket connections
- A production-ready Dockerfile and a pre-configured devcontainer for local development are provided
- GitHub Codespaces is officially supported

<details><summary>Screenshots of enhanced PWA support</summary>
<p>

**Shortcuts with native app experience:**
Bildschirmfoto 2024-08-26 um 00 45 02

</p>
</details> 

<details><summary>Screenshots of GitHub Codespaces running CodeHarbor</summary>
<p>

**Codespaces Editor:**
Bildschirmfoto 2024-08-26 um 00 43 58

**Remote Access:**
Bildschirmfoto 2024-08-26 um 00 44 00

</p>
</details> 

<details><summary>List of commits</summary>
<p>

Specifically, this PR adds the following commits:

- Upgrade to Rails 7.2 and apply new framework defaults
- File specs: Test against database.yml
- Remove left-over from pagedown-bootstrap
- Fix incorrect Rubocop reference
- Remove outdated reference to node_modules folder
- Add brakeman with GitHub action pipeline
- Add production-ready Docker image
- Add webmanifest and basic PWA support
- Apply current config for Rspec and move extensions to support directory
- Move ungrouped specs to a dedicated folder `other`
- Migrate feature specs to system specs
- Switch to Rspec native transactional fixtures to replace database cleaner
- System specs: Block test execution and wait for tubesock
- Add support for devcontainers
- Add support for GitHub Codespaces and port forwarding
- Update README with Runner Management requirement

</p>
</details>


.ruby-version
```diff
@@ -1 +1 @@
-3.3.4
+ruby-3.3.4
```
Dome-GER, line 1: We need to be a bit careful with the merge order 😇

config/initializers/assets.rb
```diff
@@ -3,7 +3,7 @@
 # Be sure to restart your server when you modify this file.
 
 # Version of your assets, change this if you want to expire all your assets.
-Rails.application.config.assets.version = '1.1'
+Rails.application.config.assets.version = '1.0'
```
Dome-GER, line 6: "Downgrade" because bootstrap explicitly required this version? I assume `1.0` is the standard value.

Dockerfile
```diff
@@ -0,0 +1,82 @@
+# syntax = docker/dockerfile:1
+
+# This Dockerfile is designed for production, not development. Use with Kamal or build'n'run by hand:
+# docker build -t my-app .
+# docker run -d -p 80:80 -p 443:443 --name my-app -e RAILS_MASTER_KEY=<value from config/master.key> my-app
+
+# Make sure RUBY_VERSION matches the Ruby version in .ruby-version
+ARG RUBY_VERSION=3.3.5
```
Dome-GER, line 8: The image is at `3.3.5` while the application in this MR is still at `3.3.4`. I totally understand why (so many PRs), but it conflicts with the comment above. ;)

docs/environment_variables.md
```diff
@@ -6,6 +6,7 @@ The following environment variables are specifically support in CodeOcean and ar
 |- |- |- |
 | `RAILS_ENV` | `development` | Specifies the Rails environment which can be configured using the files in `config/environments` |  
 | `SECRET_KEY_BASE` | ` ` | Specifies a server-side secret for Rails, must be set for production |
+| `DATABASE_URL` | ` ` | Specifies database parameters to use, rather than those specified in `config/database.yml`. An valid URL would be `postgresql://username:password@host:5432/database?pool=5` |
```
Dome-GER, line 9: "**A** valid URL would be [...]"

app/views/pwa/service-worker.js
```diff
@@ -0,0 +1,26 @@
+// Add a service worker for processing Web Push notifications:
```
Dome-GER, line 1: Has this been generated somehow? You prefer to have such "default" files added to the code base, don't you?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,3 @@
+# Make sure RUBY_VERSION matches the Ruby version in .ruby-version
+ARG RUBY_VERSION=3.3.5
```
Dome-GER, line 2: Same here ;)

.devcontainer/compose.yaml
```diff
@@ -0,0 +1,40 @@
...
+    depends_on:
+      - selenium
+      - postgres
+
+  selenium:
+    image: selenium/standalone-chromium
+    restart: unless-stopped
+
+  postgres:
+    image: postgres:16.4
```
Dome-GER, line 29: You are at PG v17 now, aren't you?

Dome-GER (approved):
Yet another exceptional PR, Sebastian! Awesome work. As usual, you are considering every possible extension and improvement.

I did not look into all documentation changes in detail, and trust your local testing.
The main conflict I see is the different Ruby and gem versions, which is the result of many open and pending PRs - but this can be fixed easily.

config/initializers/assets.rb
MrSerth: There is no dependency for the downgrade. Rather, it is shown each time an app upgrade is performed as a difference that one must skip.

Back in 2022, we increased the asset version to force a rebuild of all assets, caused by the introduction of Subresource Integrity (SRI). See 65c95a1f for details.

On our production deployment, we only keep the last five releases (and we've deployed more often since than) and also instruct browsers to cache assets for one year (which passed as well). Hence, I decided to align the version number again with the Rails defaults to make upgrades easier. It will, again, cause all assets to be rebuild. However, given the changes performed in #2497, that's actually not too bad.

Is that fine for you?

Dockerfile
MrSerth: Yes. I've bumped this commit to 3.3.5 already, because it would have been overlooked easily in https://github.com/openHPI/codeocean/pull/2517. Since https://github.com/openHPI/codeocean/pull/2517 is merged first, the commit history on `master` will be clean.

docs/environment_variables.md
MrSerth: Fixed (amended previous commit introducing the line).

.devcontainer/compose.yaml
MrSerth: Yes, we are. I've updated that -- thanks!

app/views/pwa/service-worker.js
MrSerth: Yes, this has been generated. At first, I was tempted to remove it, so that we don't keep (and deliver) this boiler-plate code to clients.

However, it is also a valuable documentation on getting started with web workers and also the default in new Rails application. The route to deliver this file exists (and is working; `pwa_service_worker`), but it is not used anywhere, so that the impact of this file is rather minimal. Still, with the aim to be up-to-date and following the Rails best practices, I then thought it might be acceptable to leave the file in the repo and use it on demand (e.g., as part of the `manifest.webmanifest`.

config/initializers/assets.rb
Dome-GER: I see, fine for me.

PR #2510, opened 2024-09-05
Extract health check for depleted environments

MrSerth:
With the extracted method it is easier to recognize the environments which caused an issue in Sentry.


mpass99 (approved):

Dome-GER (approved):
TIL: "depleted"

PR #2511, opened 2024-09-05
Gracefully handle more errors during file copy

MrSerth:
If something goes wrong while copying files, we just want to retry with a new runner. If the error is not expected (which would be only an `Runner::Error::RunnerNotFound`), we log the error to Sentry.

This PR is related to https://github.com/openHPI/poseidon/issues/649.

### DIFF
--- a/app/models/runner.rb
+++ b/app/models/runner.rb
@@ -46,7 +46,8 @@ class Runner < ApplicationRecord
   def copy_files(files)
     reserve!
     @strategy.copy_files(files)
-  rescue Runner::Error::RunnerNotFound
+  rescue Runner::Error => e
+    Sentry.capture_exception(e) unless e.is_a? Runner::Error::RunnerNotFound
     request_new_id
     save
     @strategy.copy_files(files)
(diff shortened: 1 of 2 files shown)


mpass99 (commented):
Looks good 👍 

However, we should be aware that if someone would be able to find a file copy payload returning an Internal Server Error, they could trigger an infinite loop by CodeOcean always retrying.  
On a quick check, I haven't found such payload for users. Teachers, however, are able to do so.

MrSerth:
> However, we should be aware that if someone would be able to find a file copy payload returning an Internal Server Error, they could trigger an infinite loop by CodeOcean always retrying.

I am not aware of this possibility. In the code, we retry exactly once. If the second attempt fails as well, the resulting exception is not caught and further bubbled up, where it will be handled by the corresponding controller and simply return an error.

> Teachers, however, are able to do so.

🤔 Are they? Feel free to share more details with me (also via email or Slack)...

mpass99:
> In the code, we retry exactly once. If the second attempt fails as well, the resulting exception is not caught and further bubbled up, where it will be handled by the corresponding controller and simply return an error.

I see, thanks for clarifying this mechanism.

> Are they?

They are able to create an Internal Server Error, however, I have not checked the Ruby part that this would actually lead to an Infinite loop (which as you described, does not).

mpass99 (approved):

Dome-GER (approved):

PR #2526, opened 2024-09-12
Add Solid Queue backend for Active Job

MrSerth:
With these changes, we introduce a dedicated job backend for CodeOcean and migrate away from the default `:async` backend. This will improve reliability and allow for future introduction of developer-defined jobs. For now, the backend is mainly used for library-defined jobs.

This PR assumes that the following PR is merged first:
- #2524 

However, since the PR linked is mainly about documentation, this is not a show-stopper.


Dome-GER (approved):
Looks reasonable. What's the difference between developer-defined jobs and library-defined jobs in the CodeOcean context? Is this Solid Queue wording? 🤔 Seems like I have to read more about Solid Queue soonish. 😇

MrSerth:
> What's the difference between developer-defined jobs and library-defined jobs in the CodeOcean context? Is this Solid Queue wording? 🤔 

I chose this wording myself and would define the terms as follows:

- developer-defined job: A job that is defined in this repository, namely in `app/jobs`.
- library-defined jobs: A job that is not defined in this repository, but still available to the application during run. For example, Solid Cable (see #2525) includes a `SolidCable::TrimJob` ([see here](https://github.com/rails/solid_cable/blob/7ac98418f3f713ccdc0a7c5e677d6f70be2e41ee/app/jobs/solid_cable/trim_job.rb)), that is expected to be run regularly. The default for `SolidCable::TrimJob` is once per day. Despite the newly introduced SolidCable job, there are other jobs included through the framework or library (like for ActiveStorage, or when mails are delivered later).

> Seems like I have to read more about Solid Queue soonish. 😇

Yes, that's definitely recommended! Solid Queue is now a default with Rails 8, so feel free to read more about it :+1.

PR #2538, opened 2024-09-18
Proforma: reuse existing files on import, if possible

kkoehn:
Use `xml_id_path` to recreate proforma structure with tests and model_solutions containing multiple files

also sets `xml_id_path` for exported files, if it is not set. (similiar to uuid)

### DIFF
--- a/app/services/proforma_service/convert_exercise_to_task.rb
+++ b/app/services/proforma_service/convert_exercise_to_task.rb
@@ -77,62 +77,84 @@ module ProformaService
     end
 
     def model_solutions
-      @exercise.files.filter {|file| file.role == 'reference_implementation' }.map do |file|
+      model_solutions_files = @exercise.files.filter {|file| file.role == 'reference_implementation' }.group_by {|file| xml_id_from_file(file).first }
+      model_solutions_files.map do |xml_id, files|
         ProformaXML::ModelSolution.new(
-          id: "ms-#{file.id}",
-          files: model_solution_file(file)
+          id: xml_id,
+          files: files.map {|file| model_solution_file(file) }
         )
       end
     end
 
     def model_solution_file(file)
-      [
-        task_file(file).tap do |ms_file|
-          ms_file.used_by_grader = false
-          ms_file.usage_by_lms = 'display'
-        end,
-      ]
+      task_file(file).tap do |ms_file|
+        ms_file.used_by_grader = false
+        ms_file.usage_by_lms = 'display'
+      end
     end
 
     def tests
-      @exercise.files.filter(&:teacher_defined_assessment?).map do |file|
+      @exercise.files.filter(&:teacher_defined_assessment?).group_by {|file| xml_id_from_file(file).first }.map do |xml_id, files|
         ProformaXML::Test.new(
-          id: file.id,
-          title: file.name,
-          files: test_file(file),
-          meta_data: test_meta_data(file)
+          id: xml_id,
+          title: files.first.name,
+          files: files.map {|file| test_file(file) },
+          meta_data: test_meta_data(files)
         )
       end
     end
 
-    def test_meta_data(file)
+    def xml_id_from_file(file)
+      type = if file.teacher_defined_assessment?
+               'test'
+             elsif file.reference_implementation?
+               'ms'
+             else
+               'file'
+             end
+      xml_id_path_parts = file.xml_id_path&.split('/')
+      xml_id_path  = []
+
+      xml_id_path << (xml_id_path_parts&.first || "co-#{type}-#{file.id}") unless type == 'file'
+      xml_id_path << (xml_id_path_parts&.last || file.id)
+
+      xml_id_path
+    end
+
+    def test_meta_data(files)
       {
         '@@order' => %w[test-meta-data],
         'test-meta-data' => {
-          '@@order' => %w[CodeOcean:feedback-message CodeOcean:weight CodeOcean:hidden-feedback],
+          '@@order' => %w[CodeOcean:test-file],
           '@xmlns' => {'CodeOcean' => 'codeocean.openhpi.de'},
-          'CodeOcean:feedback-message' => {
-            '@@order' => %w[$1],
-            '$1' => file.feedback_message,
-          },
-          'CodeOcean:weight' => {
-            '@@order' => %w[$1],
-            '$1' => file.weight,
-          },
-          'CodeOcean:hidden-feedback' => {
-            '@@order' => %w[$1],
-            '$1' => file.hidden_feedback,
-          },
+          'CodeOcean:test-file' => files.map do |file|
+            {
+              '@@order' => %w[CodeOcean:feedback-message CodeOcean:weight CodeOcean:hidden-feedback],
+              '@xmlns' => {'CodeOcean' => 'codeocean.openhpi.de'},
+              '@id' =>  xml_id_from_file(file).last,
+              '@name' => file.name,
+              'CodeOcean:feedback-message' => {
+                '@@order' => %w[$1],
+                '$1' => file.feedback_message,
+              },
+              'CodeOcean:weight' => {
+                '@@order' => %w[$1],
+                '$1' => file.weight,
+              },
+              'CodeOcean:hidden-feedback' => {
+                '@@order' => %w[$1],
+                '$1' => file.hidden_feedback,
+              },
+            }
+          end,
         },
       }
     end
 
     def test_file(file)
-      [
-        task_file(file).tap do |t_file|
-          t_file.used_by_grader = true
-        end,
-      ]
+      task_file(file).tap do |t_file|
+        t_file.used_by_grader = true
+      end
     end
 
     def exercise_files
@@ -168,8 +190,11 @@ module ProformaService
     end
 
     def task_file(file)
+      file.update(xml_id_path: xml_id_from_file(file).join('/')) if file.xml_id_path.blank?
+
+      xml_id = xml_id_from_file(file).last
       task_file = ProformaXML::TaskFile.new(
-        id: file.id,
+        id: xml_id,
         filename: filename(file),
         usage_by_lms: file.read_only ? 'display' : 'edit',
         visible: file.hidden ? 'no' : 'yes'
--- a/app/services/proforma_service/convert_exercise_to_task.rb
+++ b/app/services/proforma_service/convert_exercise_to_task.rb
@@ -77,8 +77,7 @@ module ProformaService
     end
 
     def model_solutions
-      model_solutions_files = @exercise.files.filter {|file| file.role == 'reference_implementation' }.group_by {|file| xml_id_from_file(file).first }
-      model_solutions_files.map do |xml_id, files|
+      @exercise.files.filter(&:reference_implementation?).group_by {|file| xml_id_from_file(file).first }.map do |xml_id, files|
         ProformaXML::ModelSolution.new(
           id: xml_id,
           files: files.map {|file| model_solution_file(file) }
--- a/app/services/proforma_service/convert_exercise_to_task.rb
+++ b/app/services/proforma_service/convert_exercise_to_task.rb
@@ -67,7 +67,7 @@ module ProformaService
 
     def proglang
       regex = %r{^openhpi/co_execenv_(?<language>[^:]*):(?<version>[^-]*)(?>-.*)?$}
-      match = regex.match @exercise.execution_environment.docker_image
+      match = regex.match @exercise&.execution_environment&.docker_image
       match ? {name: match[:language], version: match[:version]} : nil
     end
 
@@ -105,6 +105,9 @@ module ProformaService
     end
 
     def xml_id_from_file(file)
+      xml_id_path = file.xml_id_path || []
+      return xml_id_path if xml_id_path&.any?
+
       type = if file.teacher_defined_assessment?
                'test'
              elsif file.reference_implementation?
@@ -112,12 +115,11 @@ module ProformaService
              else
                'file'
              end
-      xml_id_path_parts = file.xml_id_path&.split('/')
-      xml_id_path  = []
 
-      xml_id_path << (xml_id_path_parts&.first || "co-#{type}-#{file.id}") unless type == 'file'
-      xml_id_path << (xml_id_path_parts&.last || file.id)
+      xml_id_path << "co-#{type}-#{file.id}" unless type == 'file'
+      xml_id_path << file.id.to_s
 
+      file.update!(xml_id_path: xml_id_from_file(file))
       xml_id_path
     end
 
@@ -190,8 +192,6 @@ module ProformaService
     end
 
     def task_file(file)
-      file.update(xml_id_path: xml_id_from_file(file).join('/')) if file.xml_id_path.blank?
-
       xml_id = xml_id_from_file(file).last
       task_file = ProformaXML::TaskFile.new(
         id: xml_id,
--- a/app/services/proforma_service/convert_task_to_exercise.rb
+++ b/app/services/proforma_service/convert_task_to_exercise.rb
@@ -60,44 +60,53 @@ module ProformaService
     end
 
     def files
-      model_solution_files + test_files + task_files.values.tap {|array| array.each {|file| file.role ||= 'regular_file' } }
+      model_solution_files + test_files + task_files
     end
 
     def test_files
-      @task.tests.map do |test_object|
-        task_files.delete(test_object.files.first.id).tap do |file|
-          file.weight = extract_meta_data(test_object.meta_data&.dig('test-meta-data'), 'weight').presence || 1.0
-          file.feedback_message = extract_meta_data(test_object.meta_data&.dig('test-meta-data'), 'feedback-message').presence || 'Feedback'
-          file.hidden_feedback = extract_meta_data(test_object.meta_data&.dig('test-meta-data'), 'hidden-feedback').presence || false
-          file.role ||= 'teacher_defined_test'
+      @task.tests.flat_map do |test|
+        test.files.map do |task_file|
+          codeocean_file_from_task_file(task_file, test).tap do |file|
+            file.weight = extract_meta_data(test.meta_data&.dig('test-meta-data'), 'test-file', task_file.id, 'weight').presence || 1.0
+            file.feedback_message = extract_meta_data(test.meta_data&.dig('test-meta-data'), 'test-file', task_file.id, 'feedback-message').presence || 'Feedback'
+            file.hidden_feedback = extract_meta_data(test.meta_data&.dig('test-meta-data'), 'test-file', task_file.id, 'hidden-feedback').presence || false
+            file.role = 'teacher_defined_test' unless file.teacher_defined_assessment?
+          end
         end
       end
     end
 
     def model_solution_files
-      @task.model_solutions.map do |model_solution_object|
-        task_files.delete(model_solution_object.files.first.id).tap do |file|
-          file.role ||= 'reference_implementation'
+      @task.model_solutions.flat_map do |model_solution|
+        model_solution.files.map do |task_file|
+          codeocean_file_from_task_file(task_file, model_solution).tap do |file|
+            file.role ||= 'reference_implementation'
+          end
         end
       end
     end
 
     def task_files
-      @task_files ||= @task.all_files.reject {|file| file.id == 'ms-placeholder-file' }.to_h do |task_file|
-        [task_file.id, codeocean_file_from_task_file(task_file)]
+      @task.files.reject {|file| file.id == 'ms-placeholder-file' }.map do |task_file|
+        codeocean_file_from_task_file(task_file).tap do |file|
+          file.role ||= 'regular_file'
+        end
       end
     end
 
-    def codeocean_file_from_task_file(file)
+    def codeocean_file_from_task_file(file, parent_object = nil)
       extension = File.extname(file.filename)
-      codeocean_file = CodeOcean::File.new(
+
+      codeocean_file = CodeOcean::File.where(context: @exercise).where('xml_id_path = ? OR xml_id_path LIKE ?', file.id, "%/#{file.id}").first_or_initialize(context: @exercise)
+      codeocean_file.assign_attributes(
         context: @exercise,
         file_type: file_type(extension),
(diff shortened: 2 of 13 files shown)


app/services/proforma_service/convert_exercise_to_task.rb
```diff
@@ -77,62 +77,85 @@ def uuid
     end
 
     def model_solutions
-      @exercise.files.filter {|file| file.role == 'reference_implementation' }.map do |file|
+      @exercise.files.filter {|file| file.role == 'reference_implementation' }.group_by {|file| file.xml_id_path&.split('/')&.first || "co-ms-#{file.id}" }.map do |xml_id, files|
```
MrSerth, line 80: Any chance to split this line into dedicated part to improve readability? I got what it's doing, but it seems pretty complex.

<hr>

And why is the `xml_id_from_file(file)` method not used?

app/services/proforma_service/convert_exercise_to_task.rb
```diff
@@ -77,62 +77,85 @@ def uuid
...
         ProformaXML::ModelSolution.new(
-          id: "ms-#{file.id}",
-          files: model_solution_file(file)
+          id: xml_id || "ms-#{files.first.id}",
```
MrSerth, line 82: In which occasions is `xml_id` `nil`, so that `"ms-#{files.first.id}"` is used?

app/services/proforma_service/convert_exercise_to_task.rb
```diff
@@ -77,62 +77,85 @@ def uuid
...
+                  '@@order' => %w[$1],
+                  '$1' => file.weight,
+                },
+                'CodeOcean:hidden-feedback' => {
+                  '@@order' => %w[$1],
+                  '$1' => file.hidden_feedback,
+                },
+              }
+            ]
+          end,
```
MrSerth, line 150: I have the feeling that something is odd with the meta data. In my tests, I got the following results with openHPI/CodeHarbor@da1ecdda:

```xml
<test-meta-data xmlns:CodeOcean="codeocean.openhpi.de">
  <CodeOcean:test-file/>
</test-meta-data>
```

(despite having regular data in CodeOcean)

app/services/proforma_service/convert_exercise_to_task.rb
```diff
@@ -77,62 +77,85 @@ def uuid
...
-      ]
+      task_file(file).tap do |ms_file|
+        ms_file.used_by_grader = false
+        ms_file.usage_by_lms = 'display'
+      end
     end
 
     def tests
-      @exercise.files.filter(&:teacher_defined_assessment?).map do |file|
+      @exercise.files.filter(&:teacher_defined_assessment?).group_by {|file| xml_id_from_file(file).first }.map do |xml_id, files|
```
MrSerth, line 96: Can you just assist me in understanding why you're grouping here (and for model solutions, but not files)? Is this related to the duplication we do in case of a re-used XML ID (like a file that is referenced multiple times)?

app/services/proforma_service/convert_task_to_exercise.rb
```diff
@@ -60,44 +60,56 @@ def string_to_bool(str)
...
-      @task_files ||= @task.all_files.reject {|file| file.id == 'ms-placeholder-file' }.to_h do |task_file|
-        [task_file.id, codeocean_file_from_task_file(task_file)]
+      @task.files.reject {|file| file.id == 'ms-placeholder-file' }.map do |task_file|
+        codeocean_file_from_task_file(task_file).tap do |file|
+          file.role ||= 'regular_file'
+        end
       end
+      # @task_files ||= @task.all_files.reject {|file| file.id == 'ms-placeholder-file' }.to_h do |task_file|
+      #   [task_file.id, codeocean_file_from_task_file(task_file)]
+      # end
```
MrSerth, line 97: Some leftover?

app/services/proforma_service/convert_task_to_exercise.rb
```diff
@@ -60,44 +60,56 @@ def string_to_bool(str)
...
     end
 
-    def codeocean_file_from_task_file(file)
+    def codeocean_file_from_task_file(file, parent_object = nil)
       extension = File.extname(file.filename)
-      codeocean_file = CodeOcean::File.new(
+
+      codeocean_file = CodeOcean::File.where(context: @exercise).where('xml_id_path LIKE ?', "%#{file.id}").first_or_initialize(context: @exercise)
```
MrSerth, line 103: The `LIKE` query looks dangerous. If I have a `file.id` of `1` (which is possible, since CodeOcean uses increasing numbers starting from 1 actually), any `xml_id_path` ending in `1` belonging to the current exercise is found. 

Otherwise, you could also try to use `@exercise.where(...)` to make it a little shorter.

app/services/proforma_service/convert_exercise_to_task.rb
```diff
@@ -77,62 +77,85 @@ def uuid
...
         ProformaXML::Test.new(
-          id: file.id,
-          title: file.name,
-          files: test_file(file),
-          meta_data: test_meta_data(file)
+          id: xml_id || files.first.id,
```
MrSerth, line 98: Why do model solutions require the `ms-` prefix, but tests don't (in the alternative behnd the `||`)?

MrSerth (commented):
I've done a first round of my review, during which I found a few questions. Let's check these first and I'll continue with a second round.

While reviewing, I also got one extra question: Given the changes in CodeOcean to the XML Path ID and the updated metadata, would you recommend to reexport all exercises to CodeHarbor after merging this PR?

kkoehn:
> While reviewing, I also got one extra question: Given the changes in CodeOcean to the XML Path ID and the updated metadata, would you recommend to reexport all exercises to CodeHarbor after merging this PR?

Yes that would absolutely make sense!

app/services/proforma_service/convert_exercise_to_task.rb
kkoehn: > And why is the `xml_id_from_file(file)` method not used?

it somehow escaped my refactoring :eyes:

MrSerth:
I just had one additional idea: Why aren't you using an array or other structured type for storing the XML Path ID? This would save you from splitting it at the `/` and also remove potential conflicts. Either, I would go with a `jsonb` column or (my first idea) just use an array. For the latter, we have an example with `exposed_ports` of `execution_environments`.

db/schema.rb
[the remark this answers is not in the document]
kkoehn: Done, but it's a bit unfortunate to not be able to let the mirgrations do their thing and have it result in the correct schema.

app/services/proforma_service/convert_exercise_to_task.rb
kkoehn: I'll have a look again

app/services/proforma_service/convert_task_to_exercise.rb
kkoehn: I made the query more specific, but am not sure how to use `@exercise.where` here, since I'm looking for files

kkoehn:
> I just had one additional idea: Why aren't you using an array or other structured type for storing the XML Path ID? This would save you from splitting it at the `/` and also remove potential conflicts. Either, I would go with a `jsonb` column or (my first idea) just use an array. For the latter, we have an example with `exposed_ports` of `execution_environments`.

A great idea, but I makes the query a bit convoluted.

app/services/proforma_service/convert_exercise_to_task.rb
kkoehn: Found the bug and fixed it

db/schema.rb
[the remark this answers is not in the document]
MrSerth: Yes, I totally agree. This is, unfortunately, caused due to a mismatch of the schema and the migrations. The schema file resembles the current production setup, which (as you can tell) is not the one Rails would generate. I have it on my list to fix this since quite a while...

app/services/proforma_service/convert_exercise_to_task.rb
MrSerth: Okay, I think so too -- thanks for fixing!

app/services/proforma_service/convert_exercise_to_task.rb
MrSerth: Nice, thank you! I can confirm that the meta data are being exported correctly again.

app/services/proforma_service/convert_exercise_to_task.rb
MrSerth: Okay, I see. Then, this is only "working" (or required) when reexporting a previously imported exercise. I've tested this scenario and can confirm it is working as you intended.

app/services/proforma_service/convert_task_to_exercise.rb
MrSerth: Thanks, the query looks better now. Can you add a comment to to the `where`? I was just figuring it out myself, but the syntax is not very common, I'd say:

```rb
# Find file by the given XML ID. Since the `xml_id_path` might contain multiple values (for the containing ProFormA test or model solution), the actual XML ID of the file is always the last entry in the `xml_id_path`  
```

> but am not sure how to use @exercise.where here, since I'm looking for files

Sorry, I missed one important part. The following should work: `@exercise.files.where('array_length(xml_id_path, 1) IS NOT NULL AND xml_id_path[array_length(xml_id_path, 1)] = ?', file.id).first_or_initialize(context: @exercise)`

app/services/proforma_service/convert_exercise_to_task.rb
MrSerth: That's right 😅

MrSerth:
> > I just had one additional idea: Why aren't you using an array or other structured type for storing the XML Path ID? This would save you from splitting it at the `/` and also remove potential conflicts. Either, I would go with a `jsonb` column or (my first idea) just use an array. For the latter, we have an example with `exposed_ports` of `execution_environments`.
> 
> A great idea, but makes the query a bit convoluted.

Thanks for taking the challenge and using this approach. Despite the difficult query (commented above), I think this solution is better overall.

app/services/proforma_service/convert_exercise_to_task.rb
```diff
@@ -77,62 +77,85 @@ def uuid
     end
 
     def model_solutions
-      @exercise.files.filter {|file| file.role == 'reference_implementation' }.map do |file|
+      model_solutions_files = @exercise.files.filter {|file| file.role == 'reference_implementation' }.group_by {|file| xml_id_from_file(file).first }
+      model_solutions_files.map do |xml_id, files|
```
MrSerth, line 81:
```suggestion
      @exercise.files.filter(&:reference_implementation?).group_by {|file| xml_id_from_file(file).first }.map do |xml_id, files|
```

MrSerth (commented):
Nicely done, thank you! The current code changes look good and I've understand them.

Unfortunately, I still had issues verifying the code changes. Once again, I was using the already well-known file changes from here:

- [task_845_new.zip](https://github.com/user-attachments/files/17137625/task_845_new.zip)
- [task_845_old.zip](https://github.com/user-attachments/files/17137626/task_845_old.zip)

First of all, when the new version was in **CodeHarbor** and I tried uploading the old version, I got an exception there (the other direction from old to new works). Then, even when resolving this in CodeHarbor, I still couldn't import the old version in **CodeOcean** if the new version was stored already.

The error I got was: `Validation failed: Files feedback message must be blank`. The problematic file is:

```
#<CodeOcean::File id: 67823922, content: "# actual tests here (in functions)\r\nfrom z_helpers...", context_id: 1101, context_type: "Exercise", file_id: nil, file_type_id: 8, hidden: true, name: "p51test", read_only: true, created_at: "2024-09-25 19:28:50.331358000 +0000", updated_at: "2024-09-25 19:29:34.363568000 +0000", native_file: nil, role: "regular_file", hashed_content: "70721d58fcbac5c9570e63633e778d9c", feedback_message: "Feedback", weight: nil, path: nil, file_template_id: nil, hidden_feedback: false, xml_id_path: ["63332861"]>
```

<hr>

Despite the listed issue, I also found an issue with the `proglang` method. Below is the fixed entry that also works if no `execution_environment` has been selected:

```rb
    def proglang
      regex = %r{^openhpi/co_execenv_(?<language>[^:]*):(?<version>[^-]*)(?>-.*)?$}
      match = regex.match @exercise&.execution_environment&.docker_image
      match ? {name: match[:language], version: match[:version]} : nil
    end
```

Could you include that, too, please (ideally with a test)?

kkoehn:
please note that `import_spec.rb:129` is flaky right now - will look into it tomorrow

MrSerth (commented):
I started reviewing tonight, but noticed another issue: Some files can be left unreferenced.

Specifically, consider an update form the "old" to the "new" version of task 845 (ZIP files attached above). As part of the update, I see the following query in CodeOcean:

```
  CodeOcean::File Update All (1.0ms)  UPDATE "files" SET "context_id" = $1, "context_type" = $2 WHERE ("files"."id") IN (SELECT "files"."id" FROM "files" WHERE "files"."context_id" = $3 AND "files"."context_type" = $4 AND ("files"."id" = $5 OR "files"."id" = $6) ORDER BY "files"."path" ASC, "files"."name" ASC)  [["context_id", nil], ["context_type", nil], ["context_id", 1101], ["context_type", "Exercise"], ["id", 67824363], ["id", 67824368]]
```

And indeed, as expected after this query, some files were updated to have a `context_id` and `context_type` of `NULL`. This shouldn't be the case, I think, and files no longer referenced should be deleted completely.

Otherwise, the functionality seems to work fine. I'll continue my review once the functionality is fixed.

kkoehn:
> I started reviewing tonight, but noticed another issue: Some files can be left unreferenced.
> 
> Specifically, consider an update form the "old" to the "new" version of task 845 (ZIP files attached above). As part of the update, I see the following query in CodeOcean:
> 
> ```
>   CodeOcean::File Update All (1.0ms)  UPDATE "files" SET "context_id" = $1, "context_type" = $2 WHERE ("files"."id") IN (SELECT "files"."id" FROM "files" WHERE "files"."context_id" = $3 AND "files"."context_type" = $4 AND ("files"."id" = $5 OR "files"."id" = $6) ORDER BY "files"."path" ASC, "files"."name" ASC)  [["context_id", nil], ["context_type", nil], ["context_id", 1101], ["context_type", "Exercise"], ["id", 67824363], ["id", 67824368]]
> ```
> 
> And indeed, as expected after this query, some files were updated to have a `context_id` and `context_type` of `NULL`. This shouldn't be the case, I think, and files no longer referenced should be deleted completely.
> 
> Otherwise, the functionality seems to work fine. I'll continue my review once the functionality is fixed.

sorry for the delay - should be fixed now

app/services/proforma_service/convert_exercise_to_task.rb
```diff
@@ -168,8 +190,11 @@ def task_files
     end
 
     def task_file(file)
+      file.update(xml_id_path: xml_id_from_file(file)) if file.xml_id_path.blank?
+
+      xml_id = xml_id_from_file(file).last
```
MrSerth, line 195: This seems weird. Why not encapsulate the `update` within the `xml_id_from_file` (so that it always returns the XML ID (even if it needs to be created first)?

app/services/proforma_service/convert_exercise_to_task.rb
MrSerth, line 1: We still have an issue when exercises get duplicated. 

`ExercisesController#clone` should be:
```rb
    exercise = @exercise.duplicate(public: false, token: nil, user: current_user, uuid: nil)
```

(since otherwise duplicating breaks due to the UUID uniqueness check).

<hr>

Furthermore, when duplicating an exercise, the newly introduced `xml_id_path` on files is not reset (and hence simply kept). When the XML path ID is set through a ProFormA import, this might be okay-ish, but when it is set due to a previous export, we might mingle the IDs, don't we?

app/services/proforma_service/convert_task_to_exercise.rb
```diff
@@ -60,44 +62,54 @@ def string_to_bool(str)
...
+          file.feedback_message = nil
+        end
       end
     end
 
-    def codeocean_file_from_task_file(file)
+    def codeocean_file_from_task_file(file, parent_object = nil)
       extension = File.extname(file.filename)
-      codeocean_file = CodeOcean::File.new(
+      codeocean_file = @exercise.files.where('array_length(xml_id_path, 1) IS NOT NULL AND xml_id_path[array_length(xml_id_path, 1)] = ?', file.id).first_or_initialize
```
MrSerth, line 103: Can we add a short comment that we simply check for the last element of the array? I stumbled across this twice already...

app/services/proforma_service/convert_exercise_to_task.rb
```diff
@@ -77,62 +77,84 @@ def uuid
...
+
+      type = if file.teacher_defined_assessment?
+               'test'
+             elsif file.reference_implementation?
+               'ms'
+             else
+               'file'
+             end
+
+      xml_id_path << "co-#{type}-#{file.id}" unless type == 'file'
```
MrSerth, line 118: This first part is only needed for ProFormA and CodeHarbor, right? It won't ever be looked at in CodeOcean, will it?

MrSerth (commented):
Thank you for the continuous work on this topic. When checking the import with the logs on my machine, I (still) noticed the following (845 old to new):

```
  CodeOcean::File Update All (4.7ms)  UPDATE "files" SET "context_id" = $1, "context_type" = $2 WHERE ("files"."id") IN (SELECT "files"."id" FROM "files" WHERE "files"."context_id" = $3 AND "files"."context_type" = $4 AND ("files"."id" = $5 OR "files"."id" = $6) ORDER BY "files"."path" ASC, "files"."name" ASC)  [["context_id", nil], ["context_type", nil], ["context_id", 1101], ["context_type", "Exercise"], ["id", 67825538], ["id", 67825539]]
  ↳ app/services/proforma_service/convert_task_to_exercise.rb:21:in `import_task'
  CodeOcean::File Load (0.6ms)  SELECT "files".* FROM "files" WHERE "files"."id" IN ($1, $2) ORDER BY "files"."path" ASC, "files"."name" ASC  [["id", 67825538], ["id", 67825539]]
  ↳ app/services/proforma_service/convert_task_to_exercise.rb:34:in `import_task'
  CodeOcean::File Destroy (1.0ms)  DELETE FROM "files" WHERE "files"."id" = $1  [["id", 67825538]]
  ↳ app/services/proforma_service/convert_task_to_exercise.rb:34:in `import_task'
  CodeOcean::File Destroy (0.7ms)  DELETE FROM "files" WHERE "files"."id" = $1  [["id", 67825539]]
```

Especially, I notice two things:

1. The two files are first updated and then removed (same IDs).
2. During the update, they are still unreferenced from their context.

Furthermore, the `UPDATE` query looks a bit weird (regarding the `WHERE` clause). Nevertheless, I would tend to accept this, since the functionality is working now. Two questions still remain:
- Can we switch the order to avoid the unnecessary `UPDATE`?
- Should we opt to have a dedicated transaction (or potentially a `SELECT FOR UPDATE` / lock)?

```rb
<ActiveRecord object(s)>.with_lock do
  # Stuff
end

<ActiveRecord class>.transaction do
  # Stuff
end
```

<hr>

Otherwise, please find some domain-level comments attached.

app/services/proforma_service/convert_exercise_to_task.rb
kkoehn: Fixed the first part - I can't follow the second part. If the xml_id_path is valid for one Exercise/Task context it should be valid for a copied one as well, since we are not using CodeOcean-ids for the path, but the xml_id right?

app/services/proforma_service/convert_exercise_to_task.rb
kkoehn: good idea

app/services/proforma_service/convert_exercise_to_task.rb
kkoehn: I don't think the first part is needed at all technically - I just used  an arbitrary id that made sense to me.

app/services/proforma_service/convert_exercise_to_task.rb
MrSerth: Thanks for fixing the first part 👍

Your considerations on the second part are understandable too. And indeed, since the `xml_id_path` is only checked in the context of an exercise, this shouldn't be a problem in a technical sense.

By the way: 

- Why aren't we checking the uniqueness with something like:
  ```rb
    validates :xml_id_path, uniqueness: {scope: %I[context_id context_type]}
  ```

- Why aren't we checking that the :xml_id_path is blank if `context_type != 'Exercise'` (This requires a custom validation).

That having said, I also see advantages when the `xml_id_path` stays unchanged for duplicated exercises (since it allows easier tracing of changes, ...). Maybe we simply add a single comment (to above validations or the `duplicate` method?) to explain why we decided to keep it and that it might get out-of-sync with the `id` (which is fine and intended).

app/services/proforma_service/convert_task_to_exercise.rb
```diff
@@ -60,44 +68,55 @@ def string_to_bool(str)
...
+        end
       end
     end
 
-    def codeocean_file_from_task_file(file)
+    def codeocean_file_from_task_file(file, parent_object = nil)
       extension = File.extname(file.filename)
-      codeocean_file = CodeOcean::File.new(
+      # checking the last element of xml_id_path array for file.id
+      codeocean_file = @exercise.files.where('array_length(xml_id_path, 1) IS NOT NULL AND xml_id_path[array_length(xml_id_path, 1)] = ?', file.id).first_or_initialize
```
MrSerth, line 110: This SQL query is executed per file, thus creating an N+1 query. Rather than using SQL here, it would have been more beneficial to use in-memory filtering with Ruby. That way, we would load all files at once (using `@exercise.files`) and then simply select the matching file. I would expect that this improves performance (less SQL) and potentially readability (easier access to last array element). Since we access `@exercise.files` in `destroy_old_files` already, this would even skip the first SQL query (to get all files). 

I see this part as optimization only and, given the infrequent use of the endpoint and the time spent already, would be fine to skip this optimization here.

MrSerth (commented):
Awesome! I did another round of review and only have another marginal comment. Let's get this done this week, if you like 💪.

MrSerth (approved):
To assist finishing this PR, I tested the changes locally and think they are fine domain-wise. Thanks again, Karol!

PR #2545, opened 2024-09-23
Disable Puma stdout/stderr redirection

mpass99:
Related to [codeocean-terraform!200](https://lab.xikolo.de/codeocean/codeocean-terraform/-/merge_requests/200#note_214062)


MrSerth (approved):
Thank you! Removing the settings works as expected, so that we can integrate Puma into our systemd environment easily :+1:.

PR #2562, opened 2024-10-02
Remove non-working default for ApplicationMailer

MrSerth:
The `default` setting within the ApplicationMailer is overwriting the previously-set values from `config/action_mailer.yml`. Of course, that is not intended and needs to be fixed.


Dome-GER (approved):
Argh, I stumbled across this when reviewing the upgrade. Should have added a comment to be on the safe side.

PR #2580, opened 2024-10-10
Use openHPI-managed Tubesock fork

MrSerth:
Tubesock is outdated and should be replaced with ActionCable. For now, however, we simply patch the issues we've discovered.

I've created a custom fork, included some pending contributions and made additional suggestions.


MrSerth:
We'll need to update the `Gemfile.lock` after merging https://github.com/openHPI/tubesock/pull/5

Dome-GER (approved):

PR #2594, opened 2024-10-13
Reserve runners for the whole duration of their use

MrSerth:
This commit changes how runners are reserved and released. Previously, a runner was reserved only for the actual time it was actively used and the aim was to reserve it for a time as short as possible. The previous mechanism worked, but had some drawbacks:

For example, a learner could start a submission scoring, which would perform (1) a file copy, (2) an execution of the first test file, (3) an execution of the second test file, etc. For each of these actions, the runner would be reserved and released. Now, it could happen that another run was triggered between the execution of both test files, causing issues with the remaining scoring.

In the refactored version, a runner is reserved once before initially copying files and only released after the last command finished. This ensures that no one can interrupt an execution started once, hopefully making it more robust (and faster, since less locks are required).

### DIFF
--- a/app/models/runner.rb
+++ b/app/models/runner.rb
@@ -4,6 +4,12 @@ class Runner < ApplicationRecord
   include ContributorCreation
   belongs_to :execution_environment
 
+  # Each reservation is extended by the specified duration.
+  # This allows us to extend the reservation in case of a timeout and a subsequent runner swap.
+  # By observation, we know that a runner swap usually takes 500ms on average, sometimes 700ms.
+  # This time includes requesting a new runner and copying the submission files to the new runner.
+  RESERVATION_BUFFER = 1.second
+
   before_validation :request_id
   validates :runner_id, presence: true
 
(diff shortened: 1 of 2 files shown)


app/models/runner.rb
```diff
@@ -237,4 +242,13 @@ def request_new_id
       end
     end
   end
+
+  def lock_runner_for_permitted_execution_time!
+    # We reserve the runner for the permitted execution time plus one second to ensure that the runner is not
+    # released before the execution is finished. This is especially important for the case that a command execution
+    # timed out, but subsequent commands are scheduled (i.e., when calculating a score and swapping a runner).
+    reservation = execution_environment.permitted_execution_time.seconds + 1.second
```
Dome-GER, line 250: What is the 1 second based on? Is it based on your observation / experience or "some" smallest period?

Dome-GER (approved):

app/models/runner.rb
MrSerth: It is based on my experience. Usually, swapping a runner on my local system takes around 400ms, but I've also observed times up to 700ms. In production, I expect that the value is marginally higher, due to network latency. We can also go for 1.5 or 2 seconds if you like, below would be challenging. I can also add a comment, should I?

Dome-GER: Hmm...I'd tend to add a comment on where this comes from...

Dome-GER (approved):

app/models/runner.rb
Dome-GER: Nice!

PR #2596, opened 2024-10-13
Improve file filtering for submission execution

MrSerth:
Previously, many different executions got all files despite not necessary needed. We now extend the file filtering to apply for submit actions, remote evaluations and test executions, too.

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -220,10 +220,16 @@ class Submission < ApplicationRecord
       files = collect_files
 
       case cause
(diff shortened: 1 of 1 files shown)


Dome-GER (approved):
...and you've improved the check for teacher-defined assessments as well. ;)

MrSerth:
> ...and you've implicitly changed the check for 'run' as well. ;)

Indeed, because we're now keeping teacher-defined assessments if they are visible to learners :+1:

PR #2756, opened 2025-01-29
1680: Improve error when exporting through an account_link

kkoehn:
Added a more descriptive error message for the presumably most common error.

### DIFF
--- a/app/services/exercise_service/push_external.rb
+++ b/app/services/exercise_service/push_external.rb
@@ -17,15 +17,22 @@ class ExerciseService
           request.headers['Authorization'] = "Bearer #{@codeharbor_link.api_key}"
           request.body = body
         end
+        return nil if response.success?
+        return I18n.t('exercises.export_codeharbor.not_authorized') if response.status == 401
 
-        if response.success?
-          nil
-        else
-          response.status == 401 ? I18n.t('exercises.export_codeharbor.not_authorized') : response.body
-        end
+        handle_error(message: response.body)
+      rescue Faraday::ServerError => e
+        handle_error(error: e, message: I18n.t('exercises.export_codeharbor.server_error'))
       rescue StandardError => e
-        e.message
+        handle_error(error: e)
       end
+    end
+
+    private
+
+    def handle_error(message: nil, error: nil)
+      Sentry.capture_exception(error) if error.present?
+      ERB::Util.html_escape(message || error.to_s)
     end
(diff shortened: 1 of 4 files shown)


app/services/exercise_service/push_external.rb
```diff
@@ -18,7 +18,11 @@ def execute
           request.body = body
         end
 
-        response.success? ? nil : response.body
+        if response.success?
+          nil
+        else
+          response.status == 401 ? I18n.t('exercises.export_codeharbor.not_authorized') : response.body
```
MrSerth, line 24: Please ensure to prevent XSS attacks in this (and other) responses in CodeOcean and CodeHarbor. See https://github.com/openHPI/codeharbor/issues/1680#issuecomment-2369695178 for an example (pay attention to the `<TCPSocket:(closed)>` in the screenshot). 🎤

MrSerth: Thanks for working on a fix. Unfortunately, the issue is still present for Ruby errors, as shown in my screenshot.

Screenshot 2025-02-05 at 10 06 40

My way to reproduce the issue was to set a break point in CodeHarbor and simply wait. At some point, CodeOcean will notice the timeout and render given error. With the patch below (which might be a good idea to add generally), the timeout is 5 seconds:

```diff
Subject: [PATCH] Add timeout for ExerciseService::PushExternal
---
Index: app/services/exercise_service/push_external.rb
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/app/services/exercise_service/push_external.rb b/app/services/exercise_service/push_external.rb
--- a/app/services/exercise_service/push_external.rb	(revision f334a472b29c8c3753a2950e7e9af193115e856a)
+++ b/app/services/exercise_service/push_external.rb	(date 1738746151861)
@@ -32,6 +32,9 @@
 
     def connection
       Faraday.new(url: @codeharbor_link.push_url) do |faraday|
+        faraday.options[:open_timeout] = 5
+        faraday.options[:timeout] = 5
+
         faraday.adapter Faraday.default_adapter
       end
     end
```

MrSerth: The newest commit fixes that behavior, too. Thanks!

arkirchner (approved):
Thank you for adding the default error message. LGTM

PR #2924, opened 2025-05-22
Fix: Persist factory assignments of file for RfCs

arkirchner:
Associations created by a factory need to be persisted to avoid testing against an in memory object by mistake.


MrSerth (commented):
Thanks, looks good! This was clearly a mistake, and I assume that the file reference was made to another (unrelated) submission... 🙈 

For consistency and easier identification, I would recommend changing "comment request" to "RfC"/"RfCs".

arkirchner:
@MrSerth 
Thank you for the quick review. I have updated the comment and PR description.

MrSerth:
Thanks! I was more concerned with the commit message than the PR, but having both is even better 😉

PR #2964, opened 2025-06-13
chore: Remove erb templete requirments from editor

arkirchner:
Configurations can be suppied to JavaScript with data attributes. This simplifies the JavaScript migration to Webpack.

### DIFF
--- a/app/assets/javascripts/editor/editor.js.erb
+++ b/app/assets/javascripts/editor/editor.js.erb
@@ -33,7 +33,7 @@ var CodeOceanEditor = {
 
     lastCopyText: null,
 
-    sendEvents: <%= config_file.read['codeocean_events'] ? config_file.read['codeocean_events']['enabled'] : false %>,
+    sendEvents: document.documentElement.dataset.eventsEnabled === 'true',
     eventURL: Routes.events_path(),
     fileTypeURL: Routes.file_types_path(),
 
--- a/app/assets/javascripts/editor/participantsupport.js
+++ b/app/assets/javascripts/editor/participantsupport.js
@@ -1,5 +1,5 @@
 CodeOceanEditorFlowr = {
-  isFlowrEnabled: <%= CodeOcean::Config.new(:code_ocean).read[:flowr][:enabled] %>,
+  isFlowrEnabled: document.documentElement.dataset.flowrEnabled === 'true',
   flowrResultHtml:
     '<div class="card mb-2">' +
       '<div id="{{headingId}}" role="tab" class="card-header">' +
@@ -29,7 +29,7 @@ CodeOceanEditorFlowr = {
       var stackoverflowRequests = _.map(insights, function (insight) {
         var queryParams = {
           accepted: true,
-          pagesize: <%= CodeOcean::Config.new(:code_ocean).read[:flowr][:answers_per_query] or 3 %>,
+          pagesize: Number(document.documentElement.dataset.flowrAnswersPerQuery),
           order: 'desc',
           sort: 'relevance',
           site: 'stackoverflow',
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -91,4 +91,13 @@ module ApplicationHelper
   def yes
     tag.i(nil, class: 'fa-solid fa-check')
   end
+
+  def html_data_attributes
+    {
+      'data-default-locale' => I18n.default_locale,
+      'data-events-enabled' => CodeOcean::Config.new(:code_ocean).read.dig('codeocean_events', 'enabled'),
+      'data-flowr-enabled' => CodeOcean::Config.new(:code_ocean).read.dig(:flowr, :enabled),
+      'data-flowr-answers-per-query' => CodeOcean::Config.new(:code_ocean).read[:flowranswers_per_query] || 3,
+    }
+  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -97,9 +97,9 @@ module ApplicationHelper
 
     {
       'data-default-locale' => I18n.default_locale,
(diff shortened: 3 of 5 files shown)


app/assets/javascripts/editor/editor.js.erb
MrSerth, line 1: Shouldn't this file be renamed, too? And what about the magic comment?

app/assets/javascripts/editor/participantsupport.js
```diff
@@ -1,5 +1,5 @@
 CodeOceanEditorFlowr = {
-  isFlowrEnabled: <%= CodeOcean::Config.new(:code_ocean).read[:flowr][:enabled] %>,
+  isFlowrEnabled: document.documentElement.dataset.flowrEnabled === 'true',
```
MrSerth, line 2: As I mentioned already in another PR, we don't have a linter for JS and CSS yet. Still, some common styles are appreciated.

In this file, you used `'true'`. In `app/assets/javascripts/editor/editor.js.erb`, you used `"true"`.

app/views/layouts/application.html.slim
```diff
@@ -1,5 +1,5 @@
 doctype html
-html lang=I18n.locale data-default-locale=I18n.default_locale
+html lang=I18n.locale data-default-locale=I18n.default_locale data-events-enabled=CodeOcean::Config.new(:code_ocean).read.dig('codeocean_events', 'enabled') data-flowr-enabled=CodeOcean::Config.new(:code_ocean).read.dig(:flowr, :enabled) data-flowr-answers-per-query=CodeOcean::Config.new(:code_ocean).read[:flowranswers_per_query] || 3
```
MrSerth, line 2: Can we find a smaller scope for these data attributes? I think they are not needed globally (but only for some routes).

Also, we might need to consider some better caching: With the previous solution, the value was parsed at compile time. Now, it is embedded for each page rendering.

MrSerth (commented):
I already made a few comments before you switched to the draft, so please find them below.

Thanks for working on moderninzing the assets. I think the initiative is good, but would also recommend on focussing on some aspects before opening up too many topics and PRs in parallel.

app/assets/javascripts/editor/editor.js.erb
arkirchner: > Shouldn't this file be renamed, too? And what about the magic comment?
The magic comment was added to avoid some order dependent issues. I remove after I can move these files to Webpacker JS. I don't really understand the order issue with sprockets here. My idea is that I can use explicit includes later and grantee that required JavaScript is present.

https://github.com/openHPI/codeocean/pull/2288

MrSerth: The magic comment in `app/assets/javascripts/editor/editor.js.erb` is not linked to any order dependency issues, and hence should be tackled in conjunction with the changed config processing you do in this PR.

app/views/layouts/application.html.slim
MrSerth: > How now I don't think we need any caching strategy.

I see that differently, since each request rendering the application layout would read and parse the same config at least three times.

> The configuration causes me significant trouble. I think I will rewrite it to be in memory and easy to test soon.

Can you specify the trouble, please? Testing is relatively simple using the config and there are mulitple examples where we overwrite specific config values for tests already. This alone shouldn't be the motivation to rewrite the config store. I would think there are other, more important topics right now...

MrSerth: Some simple memorization could do the trick, or at least ensuring you are not initializing the config multiple times (There is some memorization built-in already). // cc @nenock

MrSerth: Thanks!

Now, there is only an issue with the config key for `flowranswers_per_query`; that doesn't exist, does it?

MrSerth: Thanks! I just noticed another discrepancy: For `'data-events-enabled'`, you use strings, while you use symbols for `'data-flowr-enabled'` -- why?

app/helpers/application_helper.rb
```diff
@@ -91,4 +91,15 @@ def value_column(value)
   def yes
     tag.i(nil, class: 'fa-solid fa-check')
   end
+
+  def html_data_attributes
+    config = CodeOcean::Config.new(:code_ocean).read
+
+    {
+      'data-default-locale' => I18n.default_locale,
+      'data-events-enabled' => config.dig('codeocean_events', 'enabled'),
```
arkirchner, line 100:
```suggestion
      'data-events-enabled' => config.dig(:+1: codeocean_events, :enabled),
```

app/views/layouts/application.html.slim
arkirchner: I am sorry. I forgot to adjust it when I copied it from the original version.

app/views/layouts/application.html.slim
```diff
@@ -1,5 +1,5 @@
 doctype html
-html lang=I18n.locale data-default-locale=I18n.default_locale
+html lang=I18n.locale *html_data_attributes
```
MrSerth, line 2: After we got the other parts right here, I am returning to my previous question regarding the gloabl scope. 

Previously, you wrote:

> The JavaScript needs this when the page loads. I don't want to make too large changes to JavaScript at the moment.

Some of the data attributes are not needed when the page loads, but can also be parsed later, I'd say. Rather than touching this part multiple times, I would suggest to get it right now.

arkirchner:
Closing this because the complexity has exploded through the reviews.

MrSerth:
> Closing this because the complexity has exploded through the reviews.

I am really surprised by this conclusion, and think we got a good progress already. Since we all spent some time already, the migration is needed (and you chose to work on this topic), I would be in favor of finalizing it. From my side, there is a single comment left right now, specifically the scope of the attributes.

@nenock Can you help out, please?

PR #2990, opened 2025-06-30
fix: Ruby example does not use let outside of test

arkirchner:


db/seeds/fibonacci/exercise_spec_3.rb
```diff
@@ -4,7 +4,7 @@
 require './reference'
 
 describe '#fibonacci' do
-  let(:sample_count) { 32 }
+  sample_count = 32
```
MrSerth, line 7: Thanks! What about using a constant here (and moving it outside the `describe` block)?

```suggestion
  SAMPLE_COUNT = 32
```

MrSerth (commented):
Feel free to add `Regresses a0a682f` as the last line in your commit message.

Sorry for breaking this, my fault 🙈

MrSerth (approved):

PR #2997, opened 2025-07-02
Migrate .js.erb files to .js

MrSerth:
This commit reverts the changes performed in 65416934, and uses local data attributes to pass settings to the frontend.

The current solution is not ideal yet, but sufficient to continue with other important tasks.

Relates to #2964.

### DIFF
--- a/app/helpers/exercise_helper.rb
+++ b/app/helpers/exercise_helper.rb
@@ -3,7 +3,22 @@
 module ExerciseHelper
   include LtiHelper
 
+  SETTINGS = CodeOcean::Config.new(:code_ocean)
+  EVENT_SETTINGS = SETTINGS.read[:codeocean_events] || {}
+  FLOWR_SETTINGS = SETTINGS.read[:flowr] || {}
+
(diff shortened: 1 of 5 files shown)


app/helpers/exercise_helper.rb
```diff
@@ -3,7 +3,22 @@
 module ExerciseHelper
   include LtiHelper
 
+  SETTINGS = CodeOcean::Config.new(:code_ocean)
+  EVENT_SETTINGS = SETTINGS.read[:codeocean_events] || {}
```
arkirchner, line 7: NITS: Can we prefix this instance variables to make them more specific to the editor? I worried about what will happen if another helper sets an instance variable with the same name. It will be fine as long the value is the same.

MrSerth: Which instance variables are you referring to? The constants? Aren't they automatically prefixed by the helper?

```rb
ExerciseHelper::SETTINGS
```

arkirchner (approved):

PR #3008, opened 2025-07-08
test: Add system tests for exercise administration

arkirchner:
The exercise administration form is a critical pass. The test will secure the functionality with the upcoming changes.

Relates to #2922

### DIFF
--- a/spec/system/exercises_system_spec.rb
+++ b/spec/system/exercises_system_spec.rb
@@ -51,11 +51,11 @@ RSpec.describe 'Exercise creation', :js do
         find('.ProseMirror').set(description)
       end
 
-      chosen_select(Exercise.human_attribute_name(:execution_environment), ruby.name)
+      chosenjs_select_helper(Exercise.human_attribute_name(:execution_environment), ruby.name)
 
-      chosen_date_time_select(Exercise.human_attribute_name(:submission_deadline), submission_deadline)
+      chosenjs_date_time_select_helper(Exercise.human_attribute_name(:submission_deadline), submission_deadline)
 
-      chosen_date_time_select(Exercise.human_attribute_name(:late_submission_deadline), late_submission_deadline)
+      chosenjs_date_time_select_helper(Exercise.human_attribute_name(:late_submission_deadline), late_submission_deadline)
 
       check Exercise.human_attribute_name(:public)
 
@@ -64,8 +64,8 @@ RSpec.describe 'Exercise creation', :js do
       within(find_by_id('files').all('li').last) do
         fill_in CodeOcean::File.human_attribute_name(:name), with: 'main'
 
-        chosen_select(CodeOcean::File.human_attribute_name(:file_type), ruby.file_type.name)
-        chosen_select(CodeOcean::File.human_attribute_name(:role), I18n.t('code_ocean/files.roles.main_file'))
+        chosenjs_select_helper(CodeOcean::File.human_attribute_name(:file_type), ruby.file_type.name)
+        chosenjs_select_helper(CodeOcean::File.human_attribute_name(:role), I18n.t('code_ocean/files.roles.main_file'))
 
         check(CodeOcean::File.human_attribute_name(:read_only))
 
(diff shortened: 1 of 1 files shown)


spec/system/exercises_system_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+    end
+
+    it 'creates an exercise with nested data' do
+      fill_in Exercise.human_attribute_name(:title), with: title
+      fill_in Exercise.human_attribute_name(:internal_title), with: internal_title
+
+      # description
+      within('.markdown-editor__wrapper') do
+        find('.ProseMirror').set(description)
+      end
```
kkoehn, line 52: I wonder if these very specific class selectors might cause brittleness in the future if a refactoring is upcoming

spec/system/exercises_system_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+
+        check(CodeOcean::File.human_attribute_name(:read_only))
+
+        find_by_id('editor-edit').click
+        send_keys code.strip
+      end
+
+      click_button I18n.t('shared.create', model: Exercise.model_name.human)
+
+      expect(page).to have_text 'Exercise has successfully been created.'
```
kkoehn, line 78: This could also use a translation

spec/system/exercises_system_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+
+  before do
+    visit(sign_in_path)
+    fill_in('email', with: teacher.email)
+    fill_in('password', with: attributes_for(:teacher)[:password])
+    click_button(I18n.t('sessions.new.link'))
+    wait_for_ajax
+  end
+
+  context 'when an exercise created' do
```
kkoehn, line 16:
```suggestion
  context 'when an exercise is created' do
```

spec/system/exercises_system_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+
+      code.lines.each do |code_line|
+        expect(page).to have_text(code_line.delete("\n"))
+      end
+    end
+  end
+
+  context 'when an exercise is updated' do
+    let!(:exercise) { create(:fibonacci) }
+    let(:teacher) { exercise.user }
```
kkoehn, line 102:
```suggestion
    let!(:exercise) { create(:fibonacci, user: teacher) }
```

kkoehn: I would move the teacher declaration to the outer scope (and assign the user in the exercise directly)

kkoehn (commented):
look good to me in general, I added a few nit-picky comments and suggestions

spec/system/exercises_system_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+
+      click_button I18n.t('shared.update', model: Exercise.model_name.human)
+
+      expect(page).to have_text("#{Exercise.human_attribute_name(:difficulty)}\n5")
+      expect(page).to have_text('main_exercise.rb')
+      expect(page).to have_no_text(deleted_file_name)
+    end
+  end
+
+  def chosen_select(name, value)
```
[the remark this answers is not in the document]
kkoehn: I'm not sure, if the naming for these methods is intuitive (`chosen_`). I'd prefer something like `set_value_for_select/datepicker`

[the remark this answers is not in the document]
arkirchner: I did use the `chosen` prefix because that is the name for the JavaScript library that is used. I want to make clear that `select` should be used in the default case and  `chosen_select` when the JavaScript is used.

I am not happy with this helper, and I think accessibility is important. In the future I want to replace the select elements with some accessible solution (this will make the default `select` work).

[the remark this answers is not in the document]
kkoehn: I understand the reasoning and maybe I'm not familiar enough with system-specs, but I don't see anyone intuitively understanding what this method does. Part of this is probably the ambiguous lib-name `chosen`. Maybe you could either make the name(s) more clear or add a comment alternatively?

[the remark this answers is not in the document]
kkoehn: yes, this is way better. Thank you!

Issue #2715, opened 2024-12-13
RfCs: Add possibility to report abuse


MrSerth:
Similar to the course forum on openHPI, learners might add text to an RfC question or a comment that is not appropriate. So far, learners have no possibility to report abuse within CodeOcean, but would need to report it through the Helpdesk. Let's implement an abuse system, similar to the one in openHPI. This should cover RfC questions as well as comments and ideally block a text once a configurable threshold of reports was received.

The only feature we have so far already is the possibility for admins to clear an RfC question (and to edit a comment freely).

PR #3011, opened 2025-07-09
feat: Report email with LIT authorization path

arkirchner:
To act on spam reports the teacher needs to sign in through the
course to authenticate in the study group. The LTI return URL is used
to link back to the course.

Relates to https://github.com/openHPI/codeocean/issues/2715

### DIFF
--- a/app/mailers/report_mailer.rb
+++ b/app/mailers/report_mailer.rb
@@ -8,9 +8,9 @@ class ReportMailer < ApplicationMailer
     study_group = @reported_content.submission.study_group
     exercise = @reported_content.exercise
 
-    # NOTE: This implementation assumes the curse URL is static and does not vary per user.
-    # This is currently valid for all known use cases. However, in dynamic environments—such as
-    # advanced content trees in openHPI or A/B/n testing—this assumption may break.
+    # NOTE: This implementation assumes the course URL is static and does not vary per user.
+    # This is currently valid for a majority of use cases. However, in dynamic scenarios (such as
+    # content trees in openHPI used in conjunction with A/B/n testing) this assumption may no longer hold true.
     @course_url = LtiParameter.find_by(study_group:, exercise:)&.lti_parameters&.[]('launch_presentation_return_url')
 
     mail(subject: I18n.t('report_mailer.report_content.subject', content_name: @reported_content.model_name.human))


Janis4411:
I'm not very familiar with CodeOcean, but the changes make sense to me.

MrSerth (commented):
As far as I understood the changes, the config setting is global. However, the URL is consumer-specific, which this setting doesn't reflect yet. Therefore, I would either use the "crazy" approach with the LTI params or add a new attribute to each consumer.

MrSerth:
There are other consumers configured on CodeOcean (e.g., including a staging consumer for openHPI), and this feature would be the first to break the multi-tenant approach. Of course, most traffic currently originates from openHPI production.

Regarding the LTI parameter, there should be only a single instance matching all search criteria. This is ensured by a validation we have in place:

https://github.com/openHPI/codeocean/blob/445113e7996ecaeea409bac3ebda1fc31ae2c2d1/app/models/lti_parameter.rb#L9

MrSerth:
Thanks. I can confirm that the changes work as intended.

There is one implicit assumption made: The exercise URL is static and doesn't change on a per-user basis, since a random element is picked from the LTI parameters. For all use cases we currently have, this is valid, but it could potentially change in dynamic environments (or with the advanced content tree in openHPI and content A/B/n testing, where this assumption might not hold true). This would be circumvented by using the RfC author's LTI params (which, in an unlikely case, could be missing) or by adding a comment about this and changing it on demand in the future.

arkirchner:
Thank you for this information.

I think this is a solution for right now. In the future, we might need to consider a cleaner approach.

Can this be approved as is?

MrSerth:
Just to ensure there is no misunderstanding: 

An hour ago, you wrote that you've added a comment (past tense), but I cannot see any differences in this PR, any other push activity in this repo, or within Slack/email. Since your previous post did not contain any other link or information on where to look for such a change, I am currently waiting patiently for further instructions. Please let me know if I am making a mistake and overlooking something obvious.

arkirchner:
I am very sorry I overlooked that the push failed.

app/mailers/report_mailer.rb
```diff
@@ -7,6 +7,10 @@ def report_content
     @reported_content = params.fetch(:reported_content)
     study_group = @reported_content.submission.study_group
     exercise = @reported_content.exercise
+
+    # NOTE: This implementation assumes the curse URL is static and does not vary per user.
+    # This is currently valid for all known use cases. However, in dynamic environments—such as
+    # advanced content trees in openHPI or A/B/n testing—this assumption may break.
```
MrSerth, line 13: Luckily, we don't need to curse the course ;)

```suggestion
    # NOTE: This implementation assumes the course URL is static and does not vary per user.
    # This is currently valid for a majority of use cases. However, in dynamic scenarios (such as
    # content trees in openHPI used in conjunction with A/B/n testing) this assumption may no longer hold true.
```

MrSerth (commented):
Thanks for pushing your changes, I just have a minor remark originally started by the spelling mistake.

PR #3016, opened 2025-07-14
feat: Allow users to report comments

### LINKED ISSUE #2715: RfCs: Add possibility to report abuse

arkirchner:
Comments can be used to potentially harass users who request comments. A button was added to report a message as spam.

Resolves #2715


app/models/spam_report.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+class SpamReport
```
arkirchner, line 3: I extracted this object out of the mailer because the mailer logic has become quite complex.

I have done this for the following reasons:
- I wanted to test the delegation between Comment and RfC in isolation (without rendering email).
- The email feature would become too complex with this much logic (single responsibility).

Putting this into the model is a bit debatable. I think in smaller Rails this is a valid strategy.

Another approach could have been to create separate mailers for comments and RfCs. I decided against this because I think this approach keeps more of the business logic out of the mailer.

app/assets/javascripts/request_for_comments.js
```diff
@@ -116,6 +116,9 @@ $(document).on('turbo-migration:load', function () {
             <button class="action-edit btn btn-sm btn-warning">' + I18n.t('shared.edit') + '</button> \
             <button class="action-delete btn btn-sm btn-danger">' + I18n.t('shared.destroy') + '</button> \
           </div> \
+          <div class="text-warning' + (comment.reportable ? '' : ' d-none') + '"> \
```
arkirchner, line 119: Originally I wanted to add this feature by utilizing the Turbo framework. However, the changes seemed a bit large, and I want to release this soon. Therefore, I thought it was not a good time for the Turbo migration.

app/views/comments/index.json.jbuilder
```diff
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
 json.array!(@comments) do |comment|
-  json.extract! comment, :id, :user_id, :file_id, :row, :column, :text, :username, :date, :updated, :editable
+  json.extract! comment, :id, :user_id, :file_id, :row, :column, :text
```
arkirchner, line 4: Please allow me this small refactor with this PR. I think this view-related field should be set in the template and not in the Controller and Model.

app/models/spam_report.rb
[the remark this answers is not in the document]
kkoehn: Additionally I'm not sure if `models/` is the right folder for this class. AFAIK we don't have this kind of class yet, so we would have to create a new place for this.

arkirchner:
View as a learner who commented on the RfC.



View as a teacher/admin.

app/models/spam_report.rb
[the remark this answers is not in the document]
arkirchner: I have renamed **SpamReport** to **UserContentReport**. I think it makes it more clear that any kind of user is reported not only spam.

Furthermore, I investigated where to put these kind objects. I considered turning it into a form object to stick it into app/forms. But this is not a form and just would make it more difficult. I also considered if this model is a parameter object, but I think it is not because of the business logic.

I think it should remain in models for the following reasons:

1. I think extracted objects that are not common patterns like Forms, Parameters, or Policies should remain in `/app/models`. The model folder holds objects with business logic. In my opinion it is not necessary for these objects to have a persistence layer.
3. For this kind of logic objects we have no meaningful suffix. For other patterns, the suffix like UserPolicy, UserController or SignUpForm tells you where this object is placed. Any object without that should live in `/app/models/` :thinking: 

In this example of extracting a class from a Rails controller the models folder was used as well.
https://thoughtbot.com/ruby-science/extract-class.html

I am not perfectly happy with this as well, but I don't know how to solve this in a better way.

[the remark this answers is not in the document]
kkoehn: I agree that we can leave it this way for now, since it's an exception. If we start to create more classes like this I would like to see some kind of organization. Maybe a subfolder in models or something like that.

app/models/user_content_report.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+class UserContentReport
+  def initialize(reported_content:)
+    unless [Comment, RequestForComment].include?(reported_content.class)
+      raise("#{reported_content.model_name} is not configured for spam reports.")
```
kkoehn, line 6: this could also be adjusted to reflect the new class name

config/locales/de/user_content_report.yml
```diff
@@ -1,8 +1,8 @@
 ---
 de:
-  report_mailer:
+  user_content_report_mailer:
     report_content:
       authentication: Kurs-URL für die LTI-Authentifizierung.
       prolog: 'Die folgenden Inhalte wurden als unangemessen gemeldet:'
-      subject: 'Spam Report: Ein %{content_name} in CodeOcean wurde als unangemessen markiert.'
+      subject: 'Spam Report: Ein %{human_model_name} in CodeOcean wurde als unangemessen markiert.'
```
kkoehn, line 7: this could also be adjusted to reflect the new class name

config/locales/en/user_content_report.yml
```diff
@@ -1,8 +1,8 @@
 ---
 en:
-  report_mailer:
+  user_content_report_mailer:
     report_content:
       authentication: Course URL for LTI authentication.
       prolog: 'The following content has been reported as inappropriate:'
-      subject: 'Spam Report: A %{content_name} on CodeOcean has been marked as inappropriate.'
+      subject: 'Spam Report: A %{human_model_name} on CodeOcean has been marked as inappropriate.'
```
kkoehn, line 7: this could also be adjusted to reflect the new class name

kkoehn (commented):
just another small suggestion

app/models/user_content_report.rb
MrSerth, line 1: I still have some remarks on this file, such as the location (`lib`?), the definition of single-line methods, the access of the RfC in case of a submission (model-specific association?), and object-oriented design (as mentioned in the previous PR for reporting RfCs already). Nele, would you be willing to have a look here, please?

app/models/user_content_report.rb
arkirchner: I am sorry; I am not sure who I can reflect the the new class name here.

app/models/user_content_report.rb
```diff
@@ -0,0 +1,51 @@
...
+        reported_content.text
+    end
+  end
+
+  def related_request_for_comment
+    case reported_content
+      when RequestForComment
+        reported_content
+      when Comment
+        RequestForComment.find_by!(file: reported_content.file)
```
arkirchner, line 34:
```suggestion
        reported_content.request_for_comment
```

app/models/user_content_report.rb
arkirchner: I moved the object to lib for consistency (Xikolo uses lib for this business logic objects as well).

For some reason I thought that comments did not always have an RfC. Since they do, I used the association. This also allowed me to check if the user has access to the RfC in the policy (no longer allowing comments/reports on RfCs that the user is not allowed to see).

MrSerth: > I moved the object to lib for consistency (Xikolo uses lib for this business logic objects as well).

Alright, thank you. I am not saying this is the perfect place, but it feels more consistent with the remaining code (and leaves `models` for actual subclasses of `ApplicationRecord`.

> For some reason I thought that comments did not always have an RfC. Since they do, I used the association. This also allowed me to check if the user has access to the RfC in the policy (no longer allowing comments/reports on RfCs that the user is not allowed to see).

Nice, thanks for switching to the association. My original comment was not concerned with the policy and authorization at all, but this is also a good addition :+1:.

app/models/user_content_report.rb
MrSerth: Thanks. Just to persist the knowledge here, too: The previous version had a major bug, which would cause issues with multi-file RfCs:

- The `file` attribute of RfCs indicates the active file a learner was working on when requesting help through an RfC.
- The `file` attribute of a comment refers to the file the comment belongs to.

In a scenario with two or more files, the comment being reported and the active file the learner worked on could be different. In this case, the `find_by!` would raise an exception, since no file would be found.

app/models/user_content_report.rb
kkoehn: see other comment

arkirchner:
Rebased and squashed to fix conflicts with the main branch.

spec/lib/user_content_report_spec.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe UserContentReport do
+  describe '#related_request_for_comment' do
+    it 'returns the RfC when the RfC itself is reported' do
+      rfc = build_stubbed(:rfc)
```
nenock, line 8: Why `build_stubbed` instead of `build`? :thinking:

spec/lib/user_content_report_spec.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe UserContentReport do
+  describe '#related_request_for_comment' do
```
nenock, line 6: How about adding a subject here?
```suggestion
RSpec.describe UserContentReport do
  subject(report) { described_class.new(reported_content:) }

  describe '#related_request_for_comment' do
```

spec/policies/comment_policy_spec.rb
```diff
@@ -15,6 +15,25 @@
...
+  permissions :create?, :show? do
+    let(:comment) { build_stubbed(:comment) }
+
+    it 'grants access to all user types' do
+      user_types.each do |user_type|
+        expect(described_class).to permit(build_stubbed(user_type), comment)
+      end
+    end
+
+    it 'does not grant access to users who have no access to the RfC' do
```
nenock, line 27: :pray: 
```ruby
context 'without access to the RfC` do
  before do
    ...
  end

  it 'does not grant access' do
    ...
  end
end
```

spec/system/report_request_for_comment_comments_system_spec.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Report RfC comments for inappropriate content', :js do
```
nenock, line 5: Wouldn't this be sufficient:
```suggestion
RSpec.describe 'Report Comments for inappropriate content', :js do
```

app/controllers/comments_controller.rb
```diff
@@ -67,6 +61,15 @@ def destroy
...
   end
 
+  # POST /comments/1/report.json
+  def report
+    authorize!
+
+    UserContentReportMailer.with(reported_content: @comment).report_content.deliver_later
+
+    head :no_content
+  end
```
nenock, line 71: Why don't we redirect here as well (like after reporting a RfC)? :thinking:

spec/system/report_request_for_comment_comments_system_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    fill_in('password', with: attributes_for(:learner)[:password])
+    click_button(I18n.t('sessions.new.link'))
+    visit(request_for_comment_path(create(:rfc_with_comment)))
+    find('span.ace_icon').click
+  end
+
+  context 'when reporting is enabled' do
+    let(:reports_enabled) { true }
+
+    it 'allows reporting of RfCs' do
```
nenock, line 21: This seems to be a copy-paste leftover. Could you please check the whole file for those?
```suggestion
    it 'allows reporting the comment' do
```

app/controllers/comments_controller.rb
arkirchner: I was considering this. The UI for the comment is built in JavaScript. I considered recreating the `button_to` form in JavaScript but I was a bit concerned about a hacky solution with the CSRF token.

It just looked easier to do the same as the other comment operations.

spec/system/report_request_for_comment_comments_system_spec.rb
arkirchner: Sorry!

spec/system/report_request_for_comment_comments_system_spec.rb
arkirchner: I just wanted to be more specific because the comment is very unspecific. If we allow comments on other objects there would be a collision.

I have no problem rewriting this. In this case, I should change the file to `spec/system/report_comment_system_spec.rb` as well?

lib/user_content_report.rb
```diff
@@ -0,0 +1,51 @@
...
+      when RequestForComment
+        reported_content
+      when Comment
+        reported_content.request_for_comment
+    end
+  end
+
+  private
+
+  attr_reader :reported_content
```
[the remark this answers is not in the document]
arkirchner: Sorry, just habits. I am using the instance variable now.

I am doing this because I have read it in Sandi Metz's Practical Object-Oriented Design book. In my words, it makes the point:
- Instance variables can be overwritten by any method. Accessing through a method makes accidental changes more difficult.
- Refactoring becomes easier by just replacing the `attr_reader` with a method.
- No method errors are easier to understand than just having an unexpected nil.

I will use instance variables in next time!

lib/user_content_report.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def lti_parameters = LtiParameter.find_by(study_group:, exercise:)&.lti_parameters || {}
```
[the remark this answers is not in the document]
arkirchner: OK! Noted.

spec/lib/user_content_report_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    it 'returns the RfC when the RfC itself is reported' do
```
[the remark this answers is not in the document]
arkirchner: I added a context.

spec/lib/user_content_report_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      rfc = build_stubbed(:rfc)
+
+      expect(described_class.new(reported_content: rfc).related_request_for_comment).to eq rfc
+    end
+
+    it 'returns the associated RfC when a comment is reported' do
+      rfc = create(:rfc)
```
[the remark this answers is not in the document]
arkirchner: Done!

spec/lib/user_content_report_spec.rb
arkirchner: Done.

app/views/comments/index.json.jbuilder
```diff
@@ -1,6 +1,11 @@
...
+  json.username comment.user.displayname
+  json.date comment.created_at.strftime('%d.%m.%Y %k:%M')
+  json.updated(comment.created_at != comment.updated_at)
```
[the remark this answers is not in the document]
arkirchner: Both work. Going with your solution. :shrug:

spec/lib/user_content_report_spec.rb
```diff
@@ -0,0 +1,92 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe UserContentReport do
+  subject(:report) { described_class.new(reported_content:) }
+
+  describe '#related_request_for_comment' do
+    context 'when a PfC is reported' do
```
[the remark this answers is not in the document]
arkirchner: Fixed! Thanks

PR #3017, opened 2025-07-14
test: Add tests for comment policy

arkirchner:
Related to #2715

### DIFF
--- a/app/policies/comment_policy.rb
+++ b/app/policies/comment_policy.rb
@@ -9,7 +9,7 @@ class CommentPolicy < ApplicationPolicy
     everyone
   end
 
-  %i[new? destroy? update? edit?].each do |action|
+  %i[destroy? update? edit?].each do |action|
     define_method(action) { admin? || author? || teacher_in_study_group? }
   end
 
--- a/spec/policies/comment_policy_spec.rb
+++ b/spec/policies/comment_policy_spec.rb
@@ -22,11 +22,11 @@ RSpec.describe CommentPolicy do
       expect(described_class).to permit(comment.user, comment)
     end
 
-    it 'grants no access to other learners' do
+    it 'does not grant access to other learners' do
       expect(described_class).not_to permit(build_stubbed(:learner), comment)
     end
 
-    it 'grants no access to teachers not in the same study group' do
+    it 'does not grant access to teachers not in the same study group' do
       expect(described_class).not_to permit(build_stubbed(:teacher), comment)
     end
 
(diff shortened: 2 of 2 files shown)


app/policies/comment_policy.rb
```diff
@@ -9,7 +9,7 @@ def show?
     everyone
   end
 
-  %i[new? destroy? update? edit?].each do |action|
+  %i[destroy? update? edit?].each do |action|
```
arkirchner, line 12: `new` is unused and can be removed.

kkoehn (approved):

spec/policies/comment_policy_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  permissions :destroy?, :update?, :edit? do
+    let(:comment) { build_stubbed(:comment) }
+
+    it 'grants access to the author' do
+      expect(described_class).to permit(comment.user, comment)
+    end
+
+    it 'grants no access to other learners' do
```
MrSerth, line 25: Here and below: We often used "does not grant..." rather than "grants no...", but I don't have a specific preference (maybe except for consistency).

spec/policies/comment_policy_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      expect(described_class).not_to permit(build_stubbed(:learner), comment)
+    end
+
+    it 'grants no access to teachers not in study group' do
```
MrSerth, line 29:
```suggestion
    it 'does not grant access to teachers not in the same study group' do
```

spec/policies/comment_policy_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      expect(described_class).not_to permit(build_stubbed(:teacher), comment)
+    end
+
+    it 'grants access to teachers in study group' do
```
MrSerth, line 33:
```suggestion
    it 'grants access to teachers in the same study group' do
```

spec/policies/comment_policy_spec.rb
arkirchner: Right! I unified the spelling to use grants ...

MrSerth: Alright, but that is different from all other files, isn't it?

arkirchner: Ok I am using `does not grant`

MrSerth: Thanks 🙂

MrSerth (approved):

Issue #3021, opened 2025-07-16
Finish migration to Turbo.js to remove JS delay


arkirchner:
A new event was introduced to ensure that asset pipeline JavaScript is loaded before execution of JS features. This causes some delay until JS features are available. Realized issues are that confirmations from UJS are only working after a little while.

https://github.com/openHPI/codeocean/blob/main/app/javascript/turbo-migration.js

TODO:
- Move all JavaScript to Webpack to avoid order dependency issues.  
- Remove migration event.

PR #2953, opened 2025-06-06
Chore: Migrate Tubolinks to Turbo

arkirchner:
Trubolinks has been replaced by Turbo. The Turbolinks library has been achieved.

Resolves #2952

PR #3030, opened 2025-07-22
refactor: Move Sprocket JS files to Webpack (Part 2)

arkirchner:
Code editor modules are moved with minimal changes.

The editor is tested in the exercise implimentaion page. Admin exercise page and the RfC page.

Relates to #3021


app/javascript/sprocket_asset_import/editor/editor.js
```diff
@@ -389,7 +391,7 @@ var CodeOceanEditor = {
     updateEditorModeToFileTypeID: function (editor, fileTypeID) {
         var newMode = 'ace/mode/text'
 
-        $.ajax(this.fileTypeURL + '/' + fileTypeID, {
+        $.ajax(this.fileTypeURL() + '/' + fileTypeID, {
```
arkirchner, line 394: Routes.js needs the page to be rendered before execution. Calling this on execution ensures the page is present.

ibrahimkonuk (commented):
Looks good to me, but haven't tested locally

christophblessing (changes requested):
What about the failing specs? Your changes seem not functional.

MrSerth (changes requested):
On my machine, core aspects of the application are broken, such as the navigation. In addition, errors are given on the console, which reminds me of https://github.com/openHPI/codeocean/pull/2953#pullrequestreview-2918672973. 

Image

app/javascript/sprocket-asset-import/editor/editor.js
```diff
@@ -1,4 +1,7 @@
-var CodeOceanEditor = {
+import * as Routes from 'generated/routes';
```
arkirchner, line 1: I am not quite sure why Routes needs to be included here explicitly. I am a bit worried why this worked locally for me (now also failing for me).

We are setting `Routes` on the window in the `application.js` before requiring the editor. 
https://github.com/openHPI/codeocean/blob/main/app/javascript/application.js#L77 

This breaks my understanding of how JavaScript should work for me. Is Webpacker moving JS around?

arkirchner: @christophblessing 
I agree that it would be better to implicitly import everything. In this transition, this is a bit difficult because this would mean a large code change. I want to first move the JavaScript sprockets and then refactor them with more and more implicit imports. 

@ibrahimkonuk 
Thank you for linking this article. I will look deeper into how webpack orders the objects.

I want to avoid some on of solution. Implicitly importing `Routes` here and using the window version in other places might work but will cause difficult issues down the line. I will keep this as a draft while researching.

PR #3040, opened 2025-07-28
fix: Theme selection event listeners

arkirchner:
The migration event needs to be used to set the event listeners for the theme select.

### DIFF
--- a/lib/assets/javascripts/color_mode_picker.js
+++ b/lib/assets/javascripts/color_mode_picker.js
@@ -77,7 +77,7 @@ function showActiveTheme(theme, focus = false) {
     }
 }
 
-$(document).on('turbo:load', function() {
+$(document).on('turbo-migration:load', function() {
     setTheme(getPreferredTheme())
 
     showActiveTheme(getPreferredTheme())


kkoehn (approved):

MrSerth:
@arkirchner I have major concerns with this change. Can you please provide some reasoning why you changed that, which bug you're fixing therewith, and what your long-term strategy is (without using the migration event)?

From my point of view, the event listener is not bound to any Sprockets code; therefore, I don't see any necessity to wait longer for the delayed event. Furthermore, the delayed event will cause some extended, visible "white flashes" when navigating in dark mode.

kkoehn:
I can't speak for Armin but for me (locally) the appearance selector didn't always work and sometimes needed a reload. I have not seen that (yet?) with this fix. I also have not seen any of the flashes you mentioned in dark mode.

MrSerth:
Thanks for your comments. I've spent some additional time debugging, and it seems like some race condition in the previous version was, indeed, causing the buttons to be non-functional. Sorry for that, my fault.

Since the code has been moved to webpack already, I think the event can be changed back to `turbo:load`, and this seems to work constantly in my testing.

The flashing might be visible when changing from one page to another, when the included packs change. For example, when navigating from `/exercises` to `/exercises/1` (because we add the `highlight.js` bundle to the `Exercise#show` action). In my experience and my measurements, the flashing is extended when waiting for the `turbo-migration:load` rather than the `turbo:load`.

Long story short: Since the code has been moved to webpack, can we go back to `turbo:load` today?

MrSerth:
Alright, great! Thanks for the new PR 👍

PR #3045, opened 2025-07-29
chore: Enable Rubocop for migrations

arkirchner:
Rubocop gives valuable information about issues in migration. New migrations should be checked by the linter.


kkoehn (approved):
This is great, I didn't know this existed!

MrSerth:
@arkirchner Could you please apply the same changes in CodeHarbor, since the Rubocop configuration is closely set in both projects?

PR #3046, opened 2025-07-29
refactor: Files feedback_message without NULL state

### LINKED ISSUE #3014: Running an `user_defined_test` file errors due to missing `feedback_message`
In the current setup, running a `user_defined_test` (as `test` action) doesn't return any output. This is due to an implicit necessary `feedback_message`, which cannot and should not be configured for `user_defined_test` files.

An exception occurs in this line:

https://github.com/openHPI/codeocean/blob/445113e7996ecaeea409bac3ebda1fc31ae2c2d1/app/models/submission.rb#L329-L335

Most likely, `Kramdown::Document.new` doesn't accept `file.feedback_message` to be `nil`. A potential solution would be to switch the `else` to:
 ```rb
elsif output[:file_role] != 'user_defined_test'
```

### PR DESCRIPTION
**Pull request by arkirchner**
Column type string should not have a null state to avoid logical
distinction between `NULL` and `''` case.

User defined tests are created without feedback_message. Removing
the NULL case allows the new record to render the empty feedback_message.

An edge case is resolved when a test for assessment with a feedback_message
is transformed to a user defined test. The feedback_message is
overwritten, and no validation error for invisible fields is displayed.

Resolves #3014

### DIFF
--- a/spec/models/code_ocean/file_spec.rb
+++ b/spec/models/code_ocean/file_spec.rb
@@ -3,7 +3,12 @@
 require 'rails_helper'
 
 RSpec.describe CodeOcean::File do
-  let(:file) { described_class.create.tap {|file| file.update(content: nil, hidden: nil, read_only: nil) } }
+  let(:file) do
+    described_class.new.tap do |file|
+      file.assign_attributes(content: nil, hidden: nil, read_only: nil)
+      file.validate
+    end
+  end
 
   it 'validates the presence of a file type' do
     expect(file.errors[:file_type]).to be_present
@@ -11,7 +16,8 @@ RSpec.describe CodeOcean::File do
 
   it 'validates the presence of the hidden flag' do
     expect(file.errors[:hidden]).to be_present
-    file.update(hidden: false)
+    file.hidden = false
+    file.validate
     expect(file.errors[:hidden]).to be_blank
   end
 
@@ -21,19 +27,24 @@ RSpec.describe CodeOcean::File do
 
   it 'validates the presence of the read-only flag' do
     expect(file.errors[:read_only]).to be_present
-    file.update(read_only: false)
+    file.read_only = false
+    file.validate
     expect(file.errors[:read_only]).to be_blank
   end
 
   context 'with a teacher-defined test' do
-    before { file.update(role: 'teacher_defined_test') }
+    before do
+      file.role = 'teacher_defined_test'
+      file.validate
+    end
 
     it 'validates the presence of a feedback message' do
       expect(file.errors[:feedback_message]).to be_present
     end
 
     it 'validates the numericality of a weight' do
-      file.update(weight: 'heavy')
+      file.weight = 'heavy'
+      file.validate
       expect(file.errors[:weight]).to be_present
     end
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: spec/models/code_ocean/file_spec.rb [hunk 1]
```diff
@@ -3,15 +3,21 @@
 require 'rails_helper'
 
 RSpec.describe CodeOcean::File do
-  let(:file) { described_class.create.tap {|file| file.update(content: nil, hidden: nil, read_only: nil) } }
+  let(:file) do
+    described_class.new.tap do |file|
```
arkirchner, line 7: I want to make clear that the `create` does not create a record here. Therefore, I use assignment and call validate.

kkoehn:
would adding a spec to test for the bug in #3014 be an option?

MrSerth (changes requested):
I am not happy to set a blank string as a solution here. This adds further custom code (like a custom validation rather than using a built-in validation), makes analysis of the code files more difficult, and potentially increases data storage requirements.

Which other solutions to the problem did you consider? Why did you decide that way?

MrSerth:
Thanks for detailing your thoughts.

> If a column can be NULL or "", we have two different ways to represent “no content.”

I disagree. Only `NULL` represents "no content", whereas `''` represents "empty content". From a database design, that is a fundamentally difference.

> In practice, these two states often require the same handling, so it’s easier to have just one. 

This is a statement about a view, not about persistency. I see a logical differentiation between both, and an arguably "easier"  access method should not influence the data modeling.

> [various blog posts]

After reading these blog posts, I think only two of them actually address `NULL` values in the database, while the others are generally dealing with `nil` in Ruby (which might be different from the data modeling). 

> While this is not a standard that is adopted by everyone I am a big fan. This data normalization avoids conditional logic and makes the code more readable.

After careful considerations and some discussions with experienced engineers, my mind hasn't changed: Data modeling in the database should be based on the business logic, and not follow an easier access pattern for the views. Also, I want to note that we could strive for two kinds of values here, too: a present, not empty string for teacher-defined assessments and `NULL` otherwise. Therewith, we would only have two cases to deal with.

arkirchner:
> Data modeling in the database should be based on the business logic

What is the business case that requires differentiation between NULL and "". After carefully checking the code, I was unable to find one.

arkirchner:
Since there are fundamental differences between our views of what good software design is I will close this PR. I will check for the `file_role` to avoid this error.

I have avoided this solution because it adds further variants to code that could have been avoided by data normalization.

MrSerth:
> > Data modeling in the database should be based on the business logic
> 
> What is the business case that requires differentiation between NULL and "". After carefully checking the code, I was unable to find one.

So far, we don't need/require `''` at all (and, ideally, it shouldn't be in place in the database), so that this question (about a differentiation between both) does not arise. The business logic I referred to is the chosen design with different file types being stored in the same table (similar to single-table inheritance), with a missing `feedback_message` for those files were none is allowed/expected (a `main_file` should not have any feedback message, because it has no semantic meaning).

> Since there are fundamental differences between our views of what good software design is I will close this PR. I will check for the `file_role` to avoid this error.

Thank you for considering that; I'll expect a rather simple change to avoid the bug in question. Also, I am open to hearing other opinions, so in case you want to ask the team to have another look and share their perspectives, you are more than welcome to.

> I have avoided this solution because it adds further variants to code that could have been avoided by data normalization.

I am much in favor of data normalization. What about adding a migration to convert all empty strings `''` to `NULL` values and ensure no empty string is accepted any longer? This also achieves data normalization and reduces the variants to two (`NULL` or non-empty string). At the same time, this approach would follow the semantic meaning.

PR #3057, opened 2025-08-05
chore: Remove live update for study group dashboard

arkirchner:
Live updates have been disabled for 5 years. Seems safe to remove.

Related to #b2fc574


MrSerth:
Thanks for working on this part and opening a PR early. I would kindly ask not to remove the dashboard and/or the live update features. For sure, they have been disabled for about five years and were not fully functional.

Resolving these issues has been (and still is) an item on my personal follow-up list. There is a local branch with some changes lying around that I would like to finalize soon.

Disabling the live updates occurred at the beginning of the COVID-19 pandemic at a time when we just launched a [programming course on openHPI](https://open.hpi.de/courses/javaeinstieg2020). At that time, we had some performance regressions, which ultimately led to a multi-year refactoring of the code execution workflow (and thus is part of our journey towards Poseidon). Due to the difficult times back then and the stress we experienced, we aimed to identify potential issues and disabled components not completely necessary for that course.

As I know today, disabling `trigger_rfc_action_cable_from_comment` was unnecessary, and this works fine. The `trigger_working_times_action_cable` is somewhat more expensive, and could benefit from an early return if no subscription is present. For this latter part, I have already made some further improvements.

PR #3058, opened 2025-08-05
chore: Migrate to local_time for NPM support

arkirchner:
jQuery.timeago has no NPM package. local_time is used as a replacement to support the migration to Webpack.

Related to #3021


config/i18n-tasks.yml
MrSerth, line 1: I would appreciate it if you could perform only those changes needed in this file, leaving the comments in place to make adjustments when needed (and allow for an easier comparison with the defaults provided by `i18n-tasks`). Also, it seems like you removed some of the custom settings we added to enable automatic normalization.

MrSerth (commented):
An early comment for your work-in-progress. Please ignore it if you're still working on that ;)

config/i18n-tasks.yml
arkirchner: Sorry this is work in progress. I do not mean to remove this. Just was testing.

PR #3059, opened 2025-08-05
ProFormA Import: fix file duplication

kkoehn:
- adds deduplication when the same file is used multiple times (occurs when importing to CodeHarbor and the export to CodeOcean)
- adds validation for File `path`

### DIFF
--- a/spec/services/proforma_service/convert_task_to_exercise_spec.rb
+++ b/spec/services/proforma_service/convert_task_to_exercise_spec.rb
@@ -187,6 +187,27 @@ RSpec.describe ProformaService::ConvertTaskToExercise do
         expect { convert_to_exercise_service.save! }.to change(Exercise, :count).by(1)
       end
 
+      context 'with two files with similar contents' do
+        let(:file_dup) do
+          ProformaXML::TaskFile.new(
+            id: 'id2',
+            content:,
+            filename:,
+            used_by_grader: 'used_by_grader',
+            visible: 'yes',
+            usage_by_lms:,
+            binary:,
+            mimetype:
+          )
+        end
+
+        let(:files) { [file, file_dup] }
+
+        it 'creates an exercises with only one file' do
+          expect(convert_to_exercise_service.files.length).to be 1
+        end
+      end
+
       context 'when file is a Makefile' do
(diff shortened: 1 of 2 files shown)


app/models/code_ocean/file.rb
```diff
@@ -60,6 +60,8 @@ class File < ApplicationRecord
     validates :weight, absence: true, unless: :teacher_defined_assessment?
     validates :file, presence: true if :context.is_a?(Submission)
     validates :context_type, inclusion: {in: ALLOWED_CONTEXT_TYPES}
+    validates :path, uniqueness: {scope: %I[name file_type context_id context_type role]}
```
arkirchner, line 63: Don't we have records that have this issue right now? I am worried that this migration will suddenly stop teachers from editing the file.

spec/services/proforma_service/convert_task_to_exercise_spec.rb
```diff
@@ -187,6 +187,27 @@
         expect { convert_to_exercise_service.save! }.to change(Exercise, :count).by(1)
       end
 
+      context 'with two files with similar contents' do
```
arkirchner, line 190: I apologize for my lack of understanding. Should we not ensure that the exported file has no duplicates?  Won't we have the same issue in CodeHarbor after exporting the it from CodeHarbor and importing it again to CodeHarbor?

arkirchner: Are the duplicates some requirement in the ProformaXML spec?

app/models/code_ocean/file.rb
kkoehn: I'm not sure about any existing records, since I don't have access, but it would indeed be important to check before merging this PR.

spec/services/proforma_service/convert_task_to_exercise_spec.rb
kkoehn: It would be preferable to also fix the export of CodeHarbor, since I only addressed the import of tasks with multi-referenced files and not the export.

It would be even better to enable CodeHarbor to support the multi-referenced files directly (which would require a serious time investment).

PR #3065, opened 2025-08-07
Revert "test: Add system tests for exercise administration"

arkirchner:
The system test is unstable. Revert for now to allow the CI to pass.

Reverts openHPI/codeocean#3008


spec/system/exercises_system_spec.rb
```diff
@@ -1,162 +0,0 @@
-# frozen_string_literal: true
```
christophblessing, line 1: So I understand that the test is flaky.
Could we use other measures than removing it entirely? I'm thinking about retrying it or marking it as flaky.

PR #3224, opened 2026-04-16
feat: Implement API to GDPR delete users

Janis4411:
This PR introduces a possible way to GDPR delete users via a API. 

If we delete a user on openHPI we also have to delete the user on codeocean. Previously this was handled via a rake task that has to be triggered manually. 

Part of SODEV-2997

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module Api
+  class ApiController < ActionController::API
+    include ActionController::HttpAuthentication::Token::ControllerMethods
+
+    before_action :authenticate!
+
+    def authenticate!
+      authenticate_or_request_with_http_token do |token, _options|
+        ActiveSupport::SecurityUtils.secure_compare(
+          token,
+          ENV.fetch('INTERNAL_API_TOKEN', nil)
+        )
+      end
+    end
+  end
+end
--- a/app/controllers/api/internal/users_controller.rb
+++ b/app/controllers/api/internal/users_controller.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+module Api
+  module Internal
+    class UsersController < Api::ApiController
+      def destroy
+        user = ExternalUser.where(external_id: params[:id]).first
+
+        return head :not_found unless user
+
+        user.soft_delete
+        head :ok
+      end
+    end
+  end
+end
--- a/app/models/external_user.rb
+++ b/app/models/external_user.rb
@@ -8,6 +8,10 @@ class ExternalUser < User
     name.presence || "#{model_name.human} #{id}"
   end
 
+  def soft_delete
+    update!(name: 'Deleted User', email: nil, deleted_at: Time.zone.now)
+  end
+
   def webauthn_name
     "#{consumer.name}: #{displayname}"
   end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -193,6 +193,12 @@ Rails.application.routes.draw do
   mount ActionCable.server => '/cable'
   mount RailsAdmin::Engine => '/rails_admin', as: 'rails_admin'
 
+  namespace :api do
+    namespace :internal do
+      resources :users, only: :destroy
+    end
+  end
+
   # Reveal health status on /up that returns 200 if the app boots with no exceptions, otherwise 500.
   # Can be used by load balancers and uptime monitors to verify that the app is live.
   get 'up', to: 'rails/health#show', as: :rails_health_check
--- a/spec/controllers/api/api_controller_spec.rb
+++ b/spec/controllers/api/api_controller_spec.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::ApiController do
+  controller(described_class) do
+    def index
+      head :ok
+    end
+  end
+
+  before do
+    routes.draw do
+      get 'index' => 'api/api#index'
+    end
+  end
+
+  describe 'authentication' do
+    context 'with valid token' do
+      it 'allows the request' do
+        request.headers['Authorization'] = 'Bearer supersecrettoken'
+
+        get :index
+
+        expect(response).to have_http_status(:ok)
+      end
+    end
+
+    context 'with invalid token' do
+      it 'returns 401 Unauthorized' do
+        request.headers['Authorization'] = 'Bearer invalid_token'
+
+        get :index
+
+        expect(response).to have_http_status(:unauthorized)
+      end
+    end
+
+    context 'without token' do
+      it 'returns 401 Unauthorized' do
+        get :index
+
+        expect(response).to have_http_status(:unauthorized)
+      end
+    end
+
+    context 'without the INTERNAL_API_TOKEN being set' do
+      it 'returns 401 Unauthorized' do
+        allow(ENV)
+          .to receive(:fetch)
+          .with('INTERNAL_API_TOKEN', nil)
+          .and_return(nil)
+
+        request.headers['Authorization'] = "Bearer #{ENV.fetch('INTERNAL_API_TOKEN', nil)}"
+        get :index
+
+        expect(response).to have_http_status(:unauthorized)
+      end
+    end
+  end
+end
--- a/spec/request/api/internal/users/destroy_spec.rb
+++ b/spec/request/api/internal/users/destroy_spec.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'DELETE Users API', type: :request do
+  include ActiveSupport::Testing::TimeHelpers
+
+  let(:headers) { {'Authorization' => 'Bearer supersecrettoken'} }
+
+  describe 'DELETE /api/internal/users' do
+    it 'soft deletes a user' do
+      user = create(:external_user, external_id: '123456')
+
+      freeze_time
+
+      expect { delete '/api/internal/users/123456', headers: headers }
+        .to change { user.reload.deleted_at }.to(Time.zone.now)
+    end
+
+    it 'returns an error if the user is not found' do
+      delete '/api/internal/users/123456',
+        headers: headers
+      expect(response).to have_http_status(:not_found)
+    end
+  end
+end
--- a/.env
+++ b/.env
@@ -1,0 +1,1 @@
+INTERNAL_API_TOKEN=supersecrettoken
--- a/Gemfile
+++ b/Gemfile
@@ -85,6 +85,10 @@ group :development, :staging do
   gem 'slim_lint', require: false
 end
 
+group :development, :test do
+  gem 'dotenv-rails'
+end
+
 group :test do
   gem 'capybara'
   gem 'rails-controller-testing'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -150,6 +150,10 @@ GEM
     docker-api (2.4.0)
       excon (>= 0.64.0)
       multi_json
+    dotenv (3.2.0)
+    dotenv-rails (3.2.0)
+      dotenv (= 3.2.0)
+      railties (>= 6.1)
     drb (2.2.3)
     erb (5.1.1)
     erubi (1.13.1)
@@ -690,6 +694,7 @@ DEPENDENCIES
   charlock_holmes
   csv
   docker-api
+  dotenv-rails
   eventmachine
   factory_bot_rails
   faraday
@@ -810,6 +815,8 @@ CHECKSUMS
   diff-lcs (1.6.2) sha256=9ae0d2cba7d4df3075fe8cd8602a8604993efc0dfa934cff568969efb1909962
   docile (1.4.1) sha256=96159be799bfa73cdb721b840e9802126e4e03dfc26863db73647204c727f21e
   docker-api (2.4.0) sha256=824be734f4cc8718189be9c8e795b6414acbbf7e8b082a06f959a27dd8dd63e6
+  dotenv (3.2.0) sha256=e375b83121ea7ca4ce20f214740076129ab8514cd81378161f11c03853fe619d
+  dotenv-rails (3.2.0) sha256=657e25554ba622ffc95d8c4f1670286510f47f2edda9f68293c3f661b303beab
   drb (2.2.3) sha256=0b00d6fdb50995fe4a45dea13663493c841112e4068656854646f418fda13373
   erb (5.1.1) sha256=b2c26e7924551d9efbae998e17ddbef220937b6422b1d2ec7ae71417b5a1f4ec
   erubi (1.13.1) sha256=a082103b0885dbc5ecf1172fede897f9ebdb745a4b97a5e8dc63953db1ee4ad9
--- a/db/cable_schema.rb
+++ b/db/cable_schema.rb
@@ -12,7 +12,7 @@
 
 ActiveRecord::Schema[8.0].define(version: 2024_09_30_231316) do
   # These are extensions that must be enabled in order to support this database
-  enable_extension "plpgsql"
+  enable_extension "pg_catalog.plpgsql"
 
   create_table "solid_cable_messages", force: :cascade do |t|
     t.binary "channel", null: false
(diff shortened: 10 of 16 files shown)


app/controllers/api/internal/users/deletions_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+          )
+
+          head :ok
+        end
+
+        private
+
+        def authenticate_request!
+          verify_token!
+          verify_signature!
```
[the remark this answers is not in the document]
Janis4411: If I leave the return out, it could cause a `DoubleRenderError`. But rubocop complains about the redundant return.

app/controllers/api/internal/users/deletions_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+
+          unless ActiveSupport::SecurityUtils.secure_compare(
+            token.to_s,
+            ENV.fetch('OPENHPI_API_TOKEN', 'test_token')
+          )
+            return head :unauthorized
+          end
+        end
+
+        def verify_signature!
```
arkirchner: The bearer token is enough. We don't need to check the message because the bearer token also ensure this coming from a trusted source.

app/controllers/api/api_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Api
+  class ApiController < ActionController::Base
```
arkirchner: Use the API controller. Then you do not need to skip forgery protection because it is not present.

```suggestion
  class ApiController < ActionController::API
```

arkirchner: Just naming this controller application controller should be fine.

app/controllers/api/internal/users/deletions_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+        end
+
+        private
+
+        def authenticate_request!
+          verify_token!
+          verify_signature!
+        end
+
+        def verify_token!
```
arkirchner: [nits] Can you move this into the API controller? This would be nice to have for further API usage.

app/controllers/api/internal/users/deletions_controller.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+module Api
+  module Internal
+    module Users
+      class DeletionsController < Api::ApiController
+        before_action :authenticate_request!
+
+        def create
```
arkirchner: Please use standard rails actions to remove a user. 

```suggestion
    module UsersController < Api::ApiController
        before_action :authenticate_request!

        def delete
```

app/jobs/user_cleanup_job.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  private
+
+  def cleanup_user_data(user_id)
+    user = ExternalUser.find_by(external_id: user_id, consumer_id: 1) # Consumer with ID 1 is openHPI.
+
+    if user.present?
+      user.update(name: 'Deleted User', email: nil)
+    end
```
arkirchner: What we are doing here is basically is to soft delete the user. I think we can add the `soft_delete` method to ExteranlUser (alos replace it the rake task). Then we don't need to change the controller and rake task when we change it logic.

Also I think it is not complex. Let's call the soft_delete in the contorller and skip the Job.

app/models/external_user.rb
```diff
@@ -8,6 +8,10 @@ def displayname
     name.presence || "#{model_name.human} #{id}"
   end
 
+  def soft_delete!
+    update(name: 'Deleted User', email: nil)
```
arkirchner: I think it is better to blow up if the update did not work. `soft_delete` does not need a `!`. The `!` indicates a volatile version of a method. Since we do not implement two versions of this method, we should not use `!`.

```suggestion
  def soft_delete
    update!(name: 'Deleted User', email: nil)
```

app/controllers/api/internal/users/deletions_controller.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Api
+  module Internal
+    module Users
+      class DeletionsController < Api::ApiController
+        before_action :verify_token!
```
arkirchner: Can you move the before action into the application controller to make every action in the API authorized?

app/controllers/api/internal/users/deletions_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+          if user.nil?
+            return head :not_found
+          end
+
+          user.soft_delete!
+          head :ok
+        rescue StandardError => e
+          Rails.logger.error("Failed to delete user #{params[:user_id]}: #{e.message}")
+          head :internal_server_error
+        end
```
arkirchner: Why do we need this?

app/controllers/api/internal/users/deletions_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+    module Users
+      class DeletionsController < Api::ApiController
+        before_action :verify_token!
+
+        def delete
+          user = ExternalUser.where(external_id: params[:user_id], consumer_id: 1).first
+
+          if user.nil?
+            return head :not_found
+          end
```
arkirchner:
```suggestion
          user = ExternalUser.find_by(external_id: params[:user_id], consumer_id: 1)
          return head :not_found unless user
```

spec/request/api/internal/users/deletions_spec.rb
```diff
@@ -0,0 +1,101 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'DELETE Users API', type: :request do
+  include ActiveJob::TestHelper
+
```
arkirchner:
```suggestion
```

spec/request/api/internal/users/deletions_spec.rb
```diff
@@ -0,0 +1,101 @@
+# frozen_string_literal: true
```
arkirchner, line 1: Please split this test up:

1. Test that `soft_delete` works in the model test. Including that an error is raised when it does not work.
2. Create a controller test to test that a request to the API is authorized.
3. In this request, only test that the user was soft deleted. Since we don't have a method for a user to tell if a user is soft deleted, we can check if this controller removes the email from the user.

app/models/external_user.rb
arkirchner: Maybe we can add a new column `deleted_at` and add a timestamp when the user was deleted. This is good for debugging.

app/controllers/api/internal/users/deletions_controller.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Api
+  module Internal
+    module Users
+      class DeletionsController < Api::ApiController
+        before_action :verify_token!
+
+        def delete
+          user = ExternalUser.where(external_id: params[:user_id], consumer_id: 1).first
```
arkirchner: I am a bit concerned that we need to hardcode an ID here. How about adding the API token on the consumer and then loading the consumer with the token for authorization?

Then we have a current_consumer and can scope the API to that consumer.

Janis4411: This was more of an additional security check to really only delete the user from openhpi. I don't think it is possible to have a `ExternalUser` with the same `external_id` for different consumers. So we could also just omit the `consumer_id`.

`user = ExternalUser.where(external_id: params[:user_id]).first`

app/controllers/api/internal/users/deletions_controller.rb
Janis4411: You mean move it into the `Api::ApiController`, right? Yes, makes sense.

app/controllers/api/internal/users/deletions_controller.rb
Janis4411: This was my attempt so satisfy the SentryBot. In case the update action fails for whatever reason, we do not return `head :ok`

app/models/external_user.rb
Janis4411: Is the `updated_at` not serving the pretty much same purpose? Once a `ExternalUser` is updated with (name: 'Deleted User', email: nil) it won't be updated again, right?

app/controllers/api/internal/users/deletions_controller.rb
Janis4411: The `find_by` method is overwritten for users and you cannot use `external_id`.  I don't see why one would do that. That is why I went with `.where` to not have to jump through extra hoops.

app/models/user.rb:104

app/controllers/api/internal/users/deletions_controller.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Api
+  module Internal
+    module Users
+      class DeletionsController < Api::ApiController
+        def delete
```
arkirchner, line 7: Please use Rails conventions.

```suggestion
    module UsersController < Api::ApiController
        def destroy
```

app/models/external_user.rb
arkirchner: It might be. The username `Deleted User` could be used by a user. No email could be a side effect from some bug. The deleted_at just makes it clearer and is a common practice for soft deletions.

app/controllers/api/internal/users/deletions_controller.rb
arkirchner: Yes. Then we can test the application controller to only allow authorized requests with a dummy controller.

app/controllers/api/internal/users/deletions_controller.rb
arkirchner: We are raising an error now. It would return 500. This is good because we do not want to handle this case.

app/controllers/api/internal/users/deletions_controller.rb
arkirchner: OK! Let's use `where` then!

app/controllers/api/internal/users/deletions_controller.rb
arkirchner: OK. Let's do this for now.

app/controllers/api/internal/users_controller.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Api
+  module Internal
+    class UsersController < Api::ApiController
+      def delete
```
arkirchner: In Rails, the DELETE request is always called destroy.

```suggestion
      def destroy
```

config/routes.rb
```diff
@@ -193,6 +193,14 @@
   mount ActionCable.server => '/cable'
   mount RailsAdmin::Engine => '/rails_admin', as: 'rails_admin'
 
+  namespace :api do
+    namespace :internal do
+      namespace :users do
+        delete 'deleted', to: 'deletions#delete'
+      end
```
arkirchner:
```suggestion
      resources :users, only: :destroy
```

app/controllers/api/internal/users_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+        user = ExternalUser.where(external_id: params[:user_id]).first
```
arkirchner: For the user destroy action, we need to pass the ID.

```suggestion
        user = ExternalUser.where(external_id: params[:id]).first
```

app/controllers/api/internal/users_controller.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Api
+  module Internal
+    class UsersController < Api::ApiController
+      def destroy
+        user = ExternalUser.where(external_id: params[:id]).first
```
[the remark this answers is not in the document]
Janis4411: I'd say this is fine. The other consumers are not live anymore (openSAP).

app/models/external_user.rb
Janis4411: I added the `deleted_at` column and update it now in the `soft_delete` method.

app/controllers/api/api_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+
+    def verify_token!
+      token = request.headers['Authorization']&.remove('Bearer ')
+
+      unless ActiveSupport::SecurityUtils.secure_compare(
+        token.to_s,
+        ENV.fetch('OPENHPI_API_TOKEN', nil)
+      )
+        head :unauthorized
+      end
```
arkirchner: We can use some more Rails helper methods here.

```suggestion
    before_action :authenticate

    def authenticate
      authenticate_or_request_with_http_token do |token, _options|
         ActiveSupport::SecurityUtils.secure_compare(
            token,
            ENV.fetch('OPENHPI_API_TOKEN', nil)
         )
      end
    end
```

Janis4411: I get this error:

undefined method 'authenticate_or_request_with_http_token' for #<Api::Internal::UsersController:0x000000013e17c458>

It should also be available for ActionController::API, I'll have another look.

Janis4411: I had to include the `ActionController::HttpAuthentication::Token::ControllerMethods` as they are set in the  `ActionController::API` by default.

spec/controllers/api/api_controller_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      it 'returns 401 Unauthorized' do
+        request.headers['Authorization'] = 'Bearer invalid_token'
+
+        get :index
+
+        expect(response).to have_http_status(:unauthorized)
+      end
+    end
+
+    context 'without token' do
```
arkirchner, line 39: Can you add one more test where OPENHPI_API_TOKEN is not present? I want to see if no secret configuration is set and the API is still unreachable.

spec/request/api/internal/users/destroy_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      end
+
+      it 'returns 200 OK' do
+        delete "/api/internal/users/#{user_id}",
+          headers: headers
+
+        expect(response).to have_http_status(:ok)
+      end
+    end
+  end
```
arkirchner: Please keep the tests simple. Your test setup is quite complex with all these let statements. One-off record creation should be done in the test directly and not assigned as `let`. This makes test easier to understand.

```suggestion
  let(:headers) { {  'Authorization' => "Bearer #{token}" } }

  describe 'DELETE /api/internal/users' do
    it 'soft deletes a user' do
      user = create(:external_user, external_id: '123456')
      
      freeze_time
      
      expect { delete "/api/internal/users/123456", headers: headers }
          .to change { user.reload.deleted_at }.to(Time.current)
      end

      it 'returns an error if the user is not found' do
        delete "/api/internal/users/123456",
          headers: headers

        expect(response).to have_http_status(:not_found)
      end
    end
  end
```

spec/controllers/api/api_controller_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  before do
+    routes.draw do
+      get 'index' => 'api/api#index'
+    end
+  end
+
+  describe 'authentication' do
+    context 'with valid token' do
+      it 'allows the request' do
```
arkirchner, line 20: [nits] A context with only one example is not really worth it. I think a single it assertion here is clearer and more readable.

```suggestion
    it 'allows the request with valid token' do
```

spec/controllers/api/internal/users_controller_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::Internal::UsersController do
```
arkirchner: We have a request test for this API. We don't need a controller test as well.

spec/controllers/api/api_controller_spec.rb
Janis4411: But this context follows the same pattern as the other contexts.

E.g.

_context 'with valid token' do, context 'with invalid token' do, context 'without token' do_

arkirchner: I am not a fan of these contexts. I think a context should not only include one test.  A context with one test is just one individual test. This is not a must for me. I approved.

arkirchner (approved):
Please fix CI issues before merging.

Thank you very much.

